PR #10, opened 2021-10-02
Make blog more readable on my tablet

gebner:
Before and after:

bikeshedded

On a technical level this replaces the existing bootstrap-based theme by the much simpler base-jinja theme.  I've customized it so that it looks roughly like the existing theme (but I would support a redesign if there's interest).  Also all javascript dependencies except for mathjax have been removed.


PatrickMassot:
When I build locally I see
image
which is not so good.

gebner:
I forgot to commit one file, but that shouldn't make a difference for the bug that you're seeing (it just fixes a js warning).

How are you building the blog?  I don't see the bug if I run `nikola auto` on a fresh checkout in a fresh venv.

gebner:
This raised our [mobile PageSpeed score](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fleanprover-community.github.io%2Fblog%2F) by 6 points from 68 to 74.  The remaining delay is mostly mathjax.

PR #23, opened 2021-12-08
add twitter meta info

robertylewis:
Following what was done for the docs, I think this should work!


robertylewis:
It occurs to me the image says "mathlib blog" while the blog is called the "community blog." I did this without thinking and because I was basing it on the "mathlib docs" image. It could be changed with some design creativity.

eric-wieser (approved):
LGTM. Easiest just to merge and check that it works?

robertylewis:
Probably. I don't know if/how often Twitter updates those cards so someone might need to tweet to test it.

robertylewis:
The new card will look a little weird side by side with the docs image, but I think that's the lesser evil -- branding it as the community blog is better.

bryangingechen (approved):

bryangingechen:
Let's give it a shot! (Who wants to try tweeting a link?)

robertylewis:
It passes the validator: https://cards-dev.twitter.com/validator

PR #26, opened 2021-12-31
LTE update

jcommelin:


posts/lte-update.md
```diff
@@ -0,0 +1,131 @@
...
+
+In addition to these properties,
+Breen-Deligne resolutions have the favourable property that the construction yields a projective resolution `C(A)` for every `A`.
+
+For our formalization project, they also come with the significant downside that the proof of the existence of Breen-Deligne resolutions
+relies on technical results from homotopy theory,
+which haven't been formalized yet.
+
+But to our delight, it turns out that there are constructions that satisfy (1) and (2) above,
+and that are good enough for our purposes.
```
adamtopaz, line 114:
```suggestion
But to our delight, it turns out that there are related constructions that satisfy (1) and (2) above which are good enough for our purposes and which can be formalized directly.
```
The previous sentence was strange, because the B-D resolution mentioned above is such a construction to begin with.

posts/lte-update.md
```diff
@@ -0,0 +1,131 @@
...
+Indeed, there is a functorial complex known as MacLane's `Q'` construction,
+which satisfies the following result.
```
adamtopaz, line 116:
```suggestion
Indeed, the functorial complex known as "MacLane's `Q'` construction", is one such example,
which also satisfies the following result.
```

posts/lte-update.md
```diff
@@ -0,0 +1,131 @@
...
+
+In turn, these spaces of Laurent measures naturally admit a free module as submodule,
+and the quotient is the mysterious space `Mbar r' S` that appears in `first_target`.
+Together, these two short exact sequences are crucial inputs
+that will be combined with the homological tools mentioned above
+to reduce the main theorem to `first_target`.
+
+Filippo A.E. Nuccio has been working arduously on the formalization of these results,
+which amounts to Theorem 6.9 of [Analytic.pdf](https://www.math.uni-bonn.de/people/scholze/Analytic.pdf).
+
```
adamtopaz, line 95: We did a lot of work in the Fall around the various categories of comphausly/profinitely filtered gadgets, which are related to these constructions. Do you want to mention these as well? I'm not so sure, since these are objects that generic mathematicians might not really care much about.

posts/lte-update.md
```diff
@@ -0,0 +1,131 @@
...
+## Breen-Deligne resolutions and MacLane's Q'-construction
+
+Finally, to compute the `Ext`-groups, at some point projective or injective resolutions must enter the picture.
```
adamtopaz, line 98: In this context, we will use projective resolutions, so I suggest:
```suggestion
Finally, to compute the `Ext`-groups, at some point projective resolutions must enter the picture.
```

posts/lte-update.md
jcommelin: Yeah, I agree. They might turn out to be notions that people will want to use in the future. But I didn't see how to mention them in this blogpost in a clear and concise way.

PR #28, opened 2022-01-04
upsize headers

robertylewis:
This change will use `<h2>` for all top-level internal headers. `<h1>` is used for post titles.


gebner:
This took me a minute to understand.  What's going on here is that `##` in a post is apparently rendered as `<h3>`; this PR changes this to `#` which is (somewhat confusingly) rendered as `<h2>`.

PR #30, opened 2022-01-13
Correct pronouns

YaelDillies:
My pronouns are they/them. Sorry for not making it clear earlier!


robertylewis:
Oh, very sorry about that! Thanks for the correction.

PR #40, opened 2022-06-08
Add a CI script to automate "this month in mathlib"

eric-wieser:
This has to be run manually


PatrickMassot:
I'm looking at the May PR and it seems the dates are wrong. It mentions PRs from June.

PR #48, opened 2022-08-02
Add a better script for this month in mathlib

mcdoll:
I did this mainly as an excuse to look at python again. It is not really sophisticated (the parsing of the title is total nonsense), but maybe this helps 'This month in mathlib' surviving.

To run it you need to install PyGithub and python-dateutil via pip. For testing I highly recommend setting down the timeframe, github does not allow for much data to be harvested without logging into an account.


jcommelin (commented):
Thanks! This looks good to me. Maybe we should merge this with the other PR that uses CI to create a draft blogpost.

jcommelin:
Cross-ref #40

PR #55, opened 2022-09-08
Month in mathlib Aug

riccardobrasca:
I am not sure about the probability section. Also, [#16246](https://github.com/leanprover-community/mathlib/pull/16246) is maybe too technical.


PatrickMassot:
Thank you very much Riccardo! It's nice to see someone taking care of those useful and heart-warming monthly posts.

PR #84, opened 2024-07-01
Fixing two typos in May Blog.

OisinMcGuinness:
Stone-Weiertrass' fixed to 'Stone-Weierstrass'
In the "General library maintainance" section (around line 79), the title now "General library maintenance".


PatrickMassot:
Thanks!

PR #89, opened 2024-08-02
feat: add actionlint

adomani:
I find this very useful.  I simply copied the relevant file from mathlib.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,20 @@
+name: Actionlint
+on:
+  push:
+    branches:
+      - 'master'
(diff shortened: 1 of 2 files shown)


bryangingechen:
Thanks! It's certainly saved me before when working on mathlib workflows.

bryangingechen:
Ah, it seems the blog is not compatible with Python 3.12 yet: https://github.com/leanprover-community/blog/actions/runs/10218978646/job/28276284146

adomani:
Actually, I do not need python 3.12 and I think that actionlint does not require it.  I'll try with an earlier version.

PR #93, opened 2024-10-17
Add blog post on probability

RemyDegenne:


posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,241 @@
...
+This section contains pointers to the Mathlib definitions for several probability objects.
+That list might be out of date when you read this! Look around in the [documentation](https://leanprover-community.github.io/mathlib4_docs/).
+
+### CDF, pdf, variance, moments
+
+Here is a list of common concepts about real probability distributions.
+
+- Probability density function of a random variable `X : Ω → E` in a space with measure `P : Measure Ω`, with respect to measure `Q : Measure E`: `pdf X P Q`.
+- Cumulative distribution function of `P : Measure ℝ`: `cdf P`. This is a `StieltjesFunction`, a monotone right continuous real function.
+- Expectation of `X` under `P`: `P[X]`.
```
b-mehta, line 141: Could you expand on this a bit? I was trying some probability theory earlier this week, for the first time in at least a year, and the expectation notation was the only part I couldn't figure out. More specifically, have `P : Measure R` with `[MeasurableSpace R]`, and `[IsProbabilityMeasure P]`. I also have a function `R -> ENNReal`, which is measurable, and I'd like its average. I ended up using `lintegral` for this, but I expected (no pun intended) one of the expectation notations to work... 

So, could you clarify the types of `X` and `P` here, and which assumptions are needed for the notation to work?

RemyDegenne: Thanks for the feedback. All the current expectation notations correspond to Bochner integrals and not Lebesgue, so they won't work for ENNReal.

I talk about the expectation in a bit more detail earlier in the post (in the random variable section): I'll add a remark over there about the two types of integrals and the assumptions on the spaces.

RemyDegenne: I added a remark.

b-mehta: Ah perfect, thanks!

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Discrete probability
+
+In discrete probability, measurability is not an issue: every set and every function are measurable.
+The typeclass `[DiscreteMeasurableSpace Ω]` signals that every set of `Ω` is measurable and the lemma `MeasurableSet.of_discrete` provides a proof of measurability.
+To obtain measurability of a function from `Ω`, use `Measurable.of_discrete`.
+
+Any countable type with measurable singletons is a `DiscreteMeasurableSpace`. For example `ℕ` or `Fin n`.
+
+A way to define a probability measure on a discrete space `Ω` is to use the type `PMF Ω`, which stands for probability mass function.
```
b-mehta, line 117: This describes a convenient way to define a probability measure on a discrete space, but it doesn't mention if `PMF  Ω` is something that should be used as an assumption in general. For instance, let's say I want to state a theorem of the form "given a probability measure P on a discrete finite space...". Should I state the theorem over `P : PMF Ω`, or instead over `[DiscreteMeasurableSpace Ω]` with `P : Measure Ω` and `[IsProbabilityMeasure Ω]`? I'm guessing the latter, but I genuinely don't know the answer!

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+The Probability/Distributions folder of Mathlib contains several common probability laws: Exponential, Gamma, Gaussian (only in `ℝ`), Geometric, Pareto, Poisson, Uniform.
+Other measures that can be used to build probability distributions are defined elsewhere, in the MeasureTheory folder: counting measure, Dirac, Lebesgue measure.
+
+For example, the Gaussian distribution on the real line with mean `μ` and variance `v` is a `Measure ℝ`:
+```lean
+def gaussianReal (μ : ℝ) (v : ℝ≥0) : Measure ℝ := -- omitted
+```
+The definition is followed by an instance `IsProbabilityMeasure (gaussianReal μ v)` that states that this is a probability measure. It is defined as the Dirac distribution for `v = 0`.
+The file where the Gaussian is defined also contains properties of its p.d.f.
```
b-mehta, line 162:
```suggestion
The file where the Gaussian is defined also contains properties of its probability density function.
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+Two independent random variables can be defined like this:
+```lean
+variable {Ω : Type*} [MeasurableSpace Ω] {P : Measure Ω}
+  {X : Ω → ℝ} {Y : Ω → ℕ} (hX : Measurable X) (hY : Measurable Y)
+  (hXY : IndepFun X Y P)
+```
+On a measure space, we can write `IndepFun X Y` without the measure argument.
+
+For family `X : ι → Ω → ℝ` of independent random variables, use `iIndepFun`.
```
b-mehta, line 208:
```suggestion
For a family `X : ι → Ω → ℝ` of independent random variables, use `iIndepFun`.
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+For sets, use `IndepSet` (two sets) and `iIndepSet` (family of sets). For sigma-algebras, `Indep` and `iIndep`.
```
b-mehta, line 210:
```suggestion
To express independence of sets, use `IndepSet` (two sets) and `iIndepSet` (family of sets). For sigma-algebras, `Indep` and `iIndep`.
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Conditional independence
+
+The way to write that two random variables `X : Ω → E` and `Y : Ω → F` are independent conditionally on a sub-sigma-algebra `m` with respect to the measure `P` is `CondIndepFun m hm X Y P`, in which `hm : m ≤ mΩ` is a proof that `m` is a sub-sigma-algebra of the sigma-algebra `mΩ` of `Ω`. We would omit the measure argument in a `MeasureSpace`.
```
b-mehta, line 214:
```suggestion
The way to write that two random variables `X : Ω → E` and `Y : Ω → F` are conditionally independent on a sub-sigma-algebra `m` with respect to the measure `P` is `CondIndepFun m hm X Y P`, in which `hm : m ≤ mΩ` is a proof that `m` is a sub-sigma-algebra of the sigma-algebra `mΩ` of `Ω`. We would omit the measure argument in a `MeasureSpace`.
```
This reads slightly better to me

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+To write that `X : Ω → E` and `Y : Ω → F` are independent conditionally on a third random variable `Z : Ω → G` (with measurability assumption `hZ : Measurable Z`), write that they are independent given the sigma-algebra generated by `Z`.
+That is, for `mG : MeasurableSpace G` the sigma-algebra on `G`, write `CondIndepFun (mG.comap Z) hZ.comap_le X Y P`.
+As of writing this blog post, there is no shorter spelling of that fact.
+
+For families of function, use `iCondIndepFun`.
```
b-mehta, line 220:
```suggestion
For families of functions, use `iCondIndepFun`.
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+They don't work for `Y : Ω → ℝ≥0∞` since `ℝ≥0∞` is not a normed space, but those functions can be integrated with the Lebesgue integral: `∫⁻ ω, Y ω ∂P`.
+There is no expectation notation for the Lebesgue integral.
+
+## Discrete probability
+
+In discrete probability, measurability is not an issue: every set and every function are measurable.
+The typeclass `[DiscreteMeasurableSpace Ω]` signals that every set of `Ω` is measurable and the lemma `MeasurableSet.of_discrete` provides a proof of measurability.
+To obtain measurability of a function from `Ω`, use `Measurable.of_discrete`.
+
+Any countable type with measurable singletons is a `DiscreteMeasurableSpace`. For example `ℕ` or `Fin n`.
```
b-mehta, line 115:
```suggestion
Any countable type with measurable singletons is a `DiscreteMeasurableSpace`, for example `ℕ` or `Fin n`.
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+For many lemmas to apply, the set `s` will need to be a measurable set. The way to express that is `MeasurableSet s`.
+
+
+## Random variables
+
+A random variable is a measurable function from a measurable space to another.
+```lean
+variable {Ω : Type*} [MeasurableSpace Ω] {X : Ω → ℝ} (hX : Measurable X)
+```
+In that code we defined a random variable `X` from the measurable space `Ω` to `ℝ` (for which the typeclass inference system finds a measurable space instance). `hX` states that `X` is measurable, which is necessary for most manipulations.
```
b-mehta, line 96:
```suggestion
In that code we defined a random variable `X` from the measurable space `Ω` to `ℝ` (for which the typeclass inference system finds a measurable space instance). The assumption `hX` states that `X` is measurable, which is necessary for most manipulations.
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+The probability of a set `s` conditioned on another set `t` for the measure `P` is `P[s|t]`, which equals `P (s ∩ t) / P t`.
+Conditioning on `t` defines a new measure named `cond P t`, denoted by `P[|t]`. With that notation, `P[s|t] = P[|t] s`.
+
+For `X : Ω → E` and `x : E`, we write `P[|X ← x]` for the probability measure conditioned on `X = x`. It is notation for `P[|X ⁻¹' {x}]`.
+This is meaningful only if `X = x` has non-zero probability, so that definition is mostly useful for discrete probability.
+
+#### Conditional expectation
+
+The conditional expectation of a random variable `Y : Ω → F` given a sigma-algebra `m` of `Ω` is `condexp m P Y`, with notation `P[Y | m]`. This is a random variable `Ω → F`.
```
sgouezel, line 182:
```suggestion
The conditional expectation of a random variable `Y : Ω → F` given a sigma-algebra `m` of `Ω` is `condexp m P Y`, with notation `P[Y | m]`. This is a random variable `Ω → F`, and it is `m`-measurable.
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+The conditional expectation of `Y` given `X : Ω → E` can be obtained with `P[Y | mE.comap X]`, in which `mE : MeasurableSpace E` is the sigma-algebra on `E`.
+
+We have special notation for the conditional expectation of the real indicator of a set `s : Set Ω`: `P⟦s | m⟧ : Ω → ℝ`.
+
+#### Conditional probability distribution
+
+The regular conditional probability distribution of `Y : Ω → F` given `X : Ω → E`, for `F` standard Borel, is denoted by `condDistrib Y X P` (for a measure `P` on `Ω`).
+This is a Markov kernel from `E` to `F` (see further down) such that for all measurable sets `s` of `F`, for `P`-almost every `ω : Ω `, `condDistrib Y X P (X ω) s` is equal to `P⟦Y ⁻¹' s|mE.comap X⟧` (up to a mismatch in types: `ℝ≥0∞` for `condDistrib` versus `ℝ` for the conditional expectation of the indicator).
```
sgouezel, line 190:
```suggestion
This is a Markov kernel from `E` to `F` (see further down) such that for all measurable sets `s` of `F`, for `P`-almost every `ω : Ω`, `condDistrib Y X P (X ω) s` is equal to `P⟦Y ⁻¹' s|mE.comap X⟧` (up to a mismatch in types: `ℝ≥0∞` for `condDistrib` versus `ℝ` for the conditional expectation of the indicator).
```

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Random variables
+
+A random variable is a measurable function from a measurable space to another.
+```lean
+variable {Ω : Type*} [MeasurableSpace Ω] {X : Ω → ℝ} (hX : Measurable X)
+```
+In that code we defined a random variable `X` from the measurable space `Ω` to `ℝ` (for which the typeclass inference system finds a measurable space instance). `hX` states that `X` is measurable, which is necessary for most manipulations.
+
+If we define a measure `P` on `Ω`, we can talk about the law or distribution of `X`. It is by definition the map of the measure `P` by `X`, `P.map X`. There is no specific notation for that law.
```
b-mehta, line 98: The definitions `map` and `comap` are mentioned a couple of times throughout this post, I wonder if it would be useful to say something more about them? Or at least, the way in which they're useful with respect to probability theory

posts/basic-probabliity-in-lean.md
sgouezel:
```suggestion
For a family `X : ι → Ω → ℝ` of independent random variables, use `iIndepFun`.
```

b-mehta (approved):
This looks really good to me, and I have mostly only stylistic comments. There are two points where I think a bit more detail could be helpful (or at least, are things that currently confuse me when trying to use mathlib probability theory, and I'm guessing that the target audience for this includes someone like me!). But even without those, this is very valuable!

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,245 @@
...
+The conditional expectation of a random variable `Y : Ω → F` given a sigma-algebra `m` of `Ω` is `condexp m P Y`, with notation `P[Y | m]`. This is a random variable `Ω → F`.
+The conditional expectation of `Y` given `X : Ω → E` can be obtained with `P[Y | mE.comap X]`, in which `mE : MeasurableSpace E` is the sigma-algebra on `E`.
+
+We have special notation for the conditional expectation of the real indicator of a set `s : Set Ω`: `P⟦s | m⟧ : Ω → ℝ`.
+
+#### Conditional probability distribution
+
+The regular conditional probability distribution of `Y : Ω → F` given `X : Ω → E`, for `F` standard Borel, is denoted by `condDistrib Y X P` (for a measure `P` on `Ω`).
+This is a Markov kernel from `E` to `F` (see further down) such that for all measurable sets `s` of `F`, for `P`-almost every `ω : Ω `, `condDistrib Y X P (X ω) s` is equal to `P⟦Y ⁻¹' s|mE.comap X⟧` (up to a mismatch in types: `ℝ≥0∞` for `condDistrib` versus `ℝ` for the conditional expectation of the indicator).
+
```
sgouezel, line 191: Add a comment somewhere that this only makes sense in standard Borel spaces? (At other places of the blog post also, maybe?)

sgouezel (commented):
This looks great!

posts/basic-probabliity-in-lean.md
RemyDegenne: I replaced it with "conditionally independent given a sub-sigma-algebra `m`", which is the way it's spelled on wikipedia.

posts/basic-probabliity-in-lean.md
RemyDegenne: To be honest I did not use PMF much, so I am unsure about it. I wrote that it was "a way to define" and not "a convenient way to define" because I am not sure about that, but at the same time I felt like I had to include a word about PMF.

Several other repositories opted for not using PMF (because bundling the summability assumption can be inconvenient, as far as I can tell). For example https://github.com/google-deepmind/debate and https://github.com/leanprover/SampCert/tree/main.

My feeling is that stating the theorem for measures will integrate better with other things in Mathlib, but I lack the practice with PMF to be assertive about that.

posts/basic-probabliity-in-lean.md
RemyDegenne: There is "for `F` standard Borel" in the first sentence of that paragraph, so I alluded to it a bit already. Do you want me to be more explicit about it?

posts/basic-probabliity-in-lean.md
b-mehta: Indeed, this was also the choice I made in my situation! I understand the hesitation to be assertive here, but I wonder if this might be exactly the right place in which to create a convention for how these things are meant to be used

posts/basic-probabliity-in-lean.md
RemyDegenne: I added two sentences about `Measure.map` and `MeasurableSpace.comap` (which are the two definitions used here) where they are first mentioned. I don't think there is much more to say about those while staying at the introductory level of the post.
Tell me if you had something else in mind.

posts/basic-probabliity-in-lean.md
RemyDegenne: Ok let's use that opportunity to make a choice. I added a sentence about preferring Measure over PMF in lemmas.

posts/basic-probabliity-in-lean.md
RemyDegenne: I added a remark about standard Borel in the conditional independence section.

sgouezel: You read my mind: this remark came to me while I was reading the paragraph on conditional independence (where standard Borel was not mentioned), but then I thought I should mention it at the first place it was relevant, and I didn't notice it was indeed mentioned there. Sorry!

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,252 @@
...
+example (P : Measure ℝ) (s : Set ℝ) : ℝ≥0∞ := P s
+```
+The probability of the event `s` is thus `P s`.
+The type `ℝ≥0∞` represents the nonnegative reals and infinity: the measure of a set is a nonnegative real number which in general may be infinite.
+If `P` is a probability measure, it actually takes only values up to 1.
+The tactic `simp` knows that a probability measure is finite and will use the lemmas `measure_ne_top` or `measure_lt_top` to prove that `P s ≠ ∞` or `P s < ∞`.
+
+The operations on `ℝ≥0∞` are not as nicely behaved as on `ℝ`: `ℝ≥0∞` is not a ring. For example, subtraction truncates to zero.
+If one finds that lemma `lemma_name` used to transform an equation does not apply to `ℝ≥0∞`, a good thing to try is to find a lemma named like `ENNReal.lemma_name_of_something` and use that instead (it will typically require that one variable is not infinite).
+
```
sgouezel, line 86: I'm a little bit sad that we can't mention `P.real` there, that was written for PFR but never upstreamed because it was blocked by the upstreaming of the `Finiteness` tactic. @YaelDillies , didn't you mention at some point upstreaming it was in your plans?

RemyDegenne:
Since I don't know where else to put it, here is a list of things that struck me as being not ideal in today's Mathlib when writing this:
- some things are defined for one integral and not the other (expectation notation, conditional expectation). I know that Yaël has WIP code for the conditional expectation on ENNReal.
- as Sébastien wrote in a comment, we should be able to work with Real and not ENNReal for finite measures (there is code for that in PFR, needs porting and probably needs to be expanded).
- some definitions don't use `volume_tac` to use the volume measure on a `MeasureSpace`. I don't remember which, but I saw at least one while making the post.
- writing `CondIndepFun (mG.comap Z) hZ.comap_le X Y P` to state independence of `X` and `Y` given a random variable `Z` is ugly. We need a definition or notation for that case.
- our stopping times have an issue with the index type (explained in the post)
- `PMF` is not well integrated with the other parts of the library. Several people who tried to use it found it inconvenient. In general, I guess discrete probability is still underdeveloped.

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,252 @@
...
+A random variable is a measurable function from a measurable space to another.
+```lean
+variable {Ω : Type*} [MeasurableSpace Ω] {X : Ω → ℝ} (hX : Measurable X)
+```
+In that code we defined a random variable `X` from the measurable space `Ω` to `ℝ` (for which the typeclass inference system finds a measurable space instance). The assumption `hX` states that `X` is measurable, which is necessary for most manipulations.
+
+If we define a measure `P` on `Ω`, we can talk about the law or distribution of a random variable `X : Ω → E`.
+The law of `X` is a measure on `E`, with value `P (X ⁻¹' s)` on any measurable set `s` of `E`.
+This is the map of the measure `P` by `X`, `Measure.map X P` or more succinctly `P.map X`.
```
b-mehta, line 100:
```suggestion
The law of `X` is a measure on `E`, with value `P (X ⁻¹' s)` on any measurable set `s` of `E`.
This is how we define the map of the measure `P` by `X`, `Measure.map X P` or more succinctly `P.map X`.
```
optional, but I think it's helpful to say something along these lines

posts/basic-probabliity-in-lean.md
sgouezel: Definitely not! To me the blog post looks ready, in fact. Maybe it's a good idea to update it in the future when new notation or new possibilities appear, but better post it as soon as possible.

YaelDillies: You might have noticed from the queue that I have started seriously upstreaming PFR. I haven't gotten to `finiteness` yet, but it's only a matter of days.

posts/basic-probabliity-in-lean.md
```diff
@@ -0,0 +1,252 @@
...
+
+Remark: there is an issue with the current definition of stopping times, which is that if the index set of the filtration `ι` is the set of natural numbers `ℕ` then `τ` has to take values in `ℕ`. In particular it cannot be infinite.
+That can cause friction with informal stopping times which are often understood to be infinite when they are not well defined.
+
+### Transition kernels
+
+A transition kernel `κ` from a measurable space `E` to a measurable space `F` is a measurable function from `E` to `Measure F`. That is, for every measurable set `s` of `F`, the function `fun x ↦ κ x s` is measurable.
+Transition kernels are represented in Mathlib by the type `Kernel E F`.
+A kernel such that all measures in its image are probability measures is called a Markov kernel. This is denoted by `IsMarkovKernel κ`.
+There are other typeclasses for kernels that are finite (`IsFiniteKernel`) or s-finite (countable sum of finite, `IsSFiniteKernel`).
```
YaelDillies, line 250: None of these names are fully qualified. What about either fully qualifying the names or turning the code snippets into links to the docs?

PR #97, opened 2025-03-30
feat: Simprocs for the Working Mathematician

Paul-Lez:
This PR adds the first blog post in a series of three to introduce the concept of simprocs and explain how to write them.

The second blog post can be found here: #98

Co-authored-by : Yaël Dillies <yael.dillies@gmail.com>

Feedback about the content and corrections are very welcome!


bryangingechen:
Perhaps you've already considered it, but I think the two parts of this post could easily be separate posts. (Certainly most readers won't be able to make it through both parts in one sitting.)

YaelDillies:
That's definitely an option. What parts do you envision? The ones as we currently split them, or something like "How simp works" then "What are simprocs"?

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+* Performance optimisation
+* Computation
+
+### Avoiding combinatorial explosion of lemmas: The `existsAndEq` simproc
+
+The `existsAndEq` simproc is designed to simplify expressions of the form `∃ a, ... ∧ a = a' ∧ ...` where `a'` is some quantity independent of `a'` by removing the existential quantifier and replacing all occurences of `a` by `a'`.
+
+```lean
+example : ∃ (a : ℤ), a*a = 25 ∧ a = 5 := by
+  simp +arith only [existsAndEq, and_true]
```
kbuzzard, line 84: `+arith only` is (in my opinion) confusing (by which I mean: I have no idea what this does, and this is supposed to be the very first example). Can't you just run `simp` on the example and show what it turns it into? Also I am confused about why you need `and_true` given your claim about what `existsAndEq` does. Surely your argument claims that `simp [existsAndEq]` turns the goal into `5*5 = 25`? Oh -- it replaces `a = 5` by `5 = 5`?? Maybe you should explain all this clearly rather than making the user have to work it all out? I think it would be much clearer just to show the output after `simp [existsAndEq]` for pedagogical purposes rather than insisting on closing the goal.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+
+example : (∃ a, (∃ b, a + b = 5) ∧ a = 3) ↔ ∃ b, 3 + b = 5 := by
+  simp only [existsAndEq, and_true]
+```
+
+Roughly speaking, the way this metaprogram operates is a follows: whenever an expression of the form `∃ a, P a` with `P a = Q ∧ R`  is encountered:
+- Recursively traverse the expression of the predicate inside the existential quantifier to try and detect an equality `a = a'` by splitting any `∧` that is found along the way into its components.
+- If an equality is found, construct a proof that `P a` implies `a = a`.
```
kbuzzard, line 92:
```suggestion
- If an equality is found, construct a proof that `P a` implies `a = a'`.
```

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+    (∃ b, p b) = p a
+```
+
+### Performance optimisation: The `reduceIte` simproc
+
+The `reduceIte` simproc is designed to take expressions of the form `ite P a b` and replace them with `a` or `b`, depending on whether `P` can be simplified to `True` or `False` by a `simp` call.
+
+```lean
+example : ite (1 + 1 = 2) 1 2 = 1 := by
+  -- Works since `simp` can simplify `1 + 1 = 2` to `True`.
```
kbuzzard, line 105: This might be true today but it's not really `simp`s job to do this. Why not choose a better example like `0 * 37 = 0`, something which is clearly `simp`s job even for nonexperts? Maybe also explain that `simp` can do this because `zero_mul` is a simp lemma?

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+  simp only [↓reduceIte]
```
kbuzzard, line 106: What the heck is this down-arrow with no explanation whatsoever?

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+
+example (T : Type) (P : Prop) (a b : T) : ite (P ∨ ¬ P) a b = a := by
```
kbuzzard, line 108: I think it's easy to confuse `T` with `\top`. Why not use `X` instead for the type? I think part of the trick with these things is to use notation which everyone is expecting. Sure Type begins with T but actually we don't often use T for a type. alpha would also work. Use something common.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+  -- Works since `simp` can simplify `P ∨ ¬ P` to `True`.
+  simp only [↓reduceIte]
+
+example : ite FermatLastTheorem 1 2 = 1 := by
+  --This fails because `simp` can't reduce `FermatLastTheorem` to `True`
+  simp only [↓reduceIte]
```
kbuzzard, line 114: Display the goal afterwards in a comment. (beginning with the turnstile)

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+```
+
+Internally, this simproc is a small metaprogram does the following whenever an expression of the form `ite P a b` is encountered:
+- Call `simp` on `P` to get a simplified expression `P'` and a proof `r` that `P = P'`.
```
kbuzzard, line 118: r is the name for a relation, not a proof. Call it `h` or `hP`?

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+  let_expr f@ite α c i tb eb ← e | return .continue
+  let r ← simp c
+  if r.expr.isTrue then
+    let pr    := mkApp (mkApp5 (mkConst ``ite_cond_eq_true f.constLevels!) α c i tb eb) (← r.getProof)
+    return .visit { expr := tb, proof? := pr }
+  if r.expr.isFalse then
+    let pr    := mkApp (mkApp5 (mkConst ``ite_cond_eq_false f.constLevels!) α c i tb eb) (← r.getProof)
+    return .visit { expr := eb, proof? := pr }
+  return .continue
+```
```
kbuzzard, line 136: This is really intimidating. Do you really want this here, so early? Until now the post has been easy-going. It's fine to have "accelerated lecture style" with code like this later on, but the sooner you put some blob of incomprehensible code the sooner you lose readers. If you're trying to tell us something with this code, can you tell it us without any meta code at all? You claimed earlier that the first section was purely informal.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+
+Intuitively, one can think of this as a "parametric" lemma that allows one to vary the right hand side depending on the value of the left hand side.
+In the case of `reduceIte`, this allows us to combine `ite_cond_eq_true` and `ite_cond_eq_false` into a single procedure that `simp` can call.
+Notice that on can use `ite_cond_eq_true` (resp `ite_cond_eq_false`) instead of `↓reduceIte`, at the cost of introducing more steps in the simplification procedure:
+
+```
+set_option trace.Meta.Tactic.simp true in
+example : ite (1≠1) (2^3 * 0 * 50 * 50) (3^2) = 3^2 := by
```
kbuzzard, line 144: There is surely no point overcomplicating things with 2^3*0 etc etc.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+  -- 6 steps: 5 rewrites and one goal discharge
+  simp only [ite_cond_eq_false, ne_eq, not_true_eq_false]
+
+set_option trace.Meta.Tactic.simp true in
+example : ite (1≠1) (2^3 * 0 * 50 * 50) (3^2) = 3^2 := by
+  -- 4 rewrites
+  simp
```
kbuzzard, line 151: What point are you trying to make here? 4 rewrites sounds absolutely fine to me. You seem to be presenting quite a coherent argument that the simproc is not really necessary. Is this the point you want to make? What is the point of this code? Are you trying to tell us that 4 rewrites is too many? Can you give a better example where the simproc is clearly saving you work? This seems to be quite a poor justification.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+```
+
+### Computation: The `reduceDvd` simproc
+
+The `reduceDvd` simproc is designed to take expressions of the form `a | b` where `a, b` are natural number literals, and simplify them to `True` or `False`.
+
+```lean
+example : 3 ∣ 9 := by
+  simp only [Nat.reduceDvd]
```
kbuzzard, line 160: Is it called `reduceDvd` or `Nat.reduceDvd`? Be consistent? Beginners might be confused about whether these are the same thing.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+
+example : ¬ 2 ∣ 49 := by
+  simp only [Nat.reduceDvd, not_false_iff]
```
kbuzzard, line 163: Again I think `not_false_iff` is just adding to the confusion. Why not just show the goal after `simp only [Nat.reduceDvd]`? This isn't mathlib. You don't have to solve the goals. Let it simplify to `\not False` and then call `simp` again, and then say "or you can do it all in one go with `simp only [Nat.reduceDvd, not_false_iff]`". Or does `simp` work by itself? If so then say this now. Teach in small chunks.  This line is doing too much -- it's assuming a lot of familiarity which most people won't have.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+```lean
+dsimproc_decl myDSimproc (theExprToMatch _ _) := fun e ↦ do
+  write_dsimproc_here
+```
+
+To add `mySimproc`/`myDSimproc` to the standard simp set, replace `simproc_decl`/`dsimproc_decl` by `simproc`/`dsimproc`.
+
+To turn `mySimproc`/`myDSimproc` into a preprocedure (recall that postprocedure is the default), do
+```lean
+simproc_decl ↓ mySimproc (theExprToMatch _ _) := fun e ↦ do
```
Paul-Lez, line 365: @YaelDillies Doesn't this only work with `simproc` rather than `simproc_decl`?

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+
+```lean
+builtin_simproc [simp, seval] reduceDvd ((_ : Nat) ∣ _) := fun e => do
+  let_expr Dvd.dvd _ i a b ← e | return .continue
+  let some va ← fromExpr? a | return .continue
+  let some vb ← fromExpr? b | return .continue
+  if vb % va == 0 then
+    return .done { expr := mkConst ``True, proof? := mkApp3 (mkConst ``Nat.dvd_eq_true_of_mod_eq_zero) a b reflBoolTrue}
+  else
+    return .done { expr := mkConst ``False, proof? := mkApp3 (mkConst ``Nat.dvd_eq_false_of_mod_ne_zero) a b reflBoolTrue}
```
kbuzzard, line 184: Again I'm super-dubious about you dumping 6 lines of meta code with absolutely 0 explanation of anything. At the very least each line should be commented. Even better: all the meta code so far should be moved into a section after this one which starts "now let's take a look at the meta code which actually does this. Warning: in this section we will assume some experience with metaprogramming". Perhaps at the end of the easy section you can give a big list of current simprocs saying what they do.

kbuzzard (commented):
I've only read the first half of the post but it shows a lot of promise -- thanks! I left a ton of pernickety comments. I think that you should go a bit slower so more people can get more from the post. For example I think the first section should be simple enough that it can teach something to people who barely know what `simp` is, don't know anything about metaprogramming, and haven't memorised a bunch of standard logic lemmas.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+description: 'Brief introduction to simprocs and what they are useful for'
+has_math: true
+link: ''
+slug: what-is-a-simproc
+tags: 'simp, simproc, meta'
+title: What is a simproc?
+type: text
+---
+
+Lean core recently added support for custom simplification procedures, aka *simprocs*.
```
b-mehta, line 14:
```suggestion
Lean core recently added support for custom simplification procedures, aka *simprocs*.
```
Is "recent" true? I thought these have been around for quite a while, at least a year.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+which yields
+```
+[Meta.Tactic.simp.rewrite] Nat.fib_zero:1000:
+      Nat.fib 0
+    ==>
+      0
+[Meta.Tactic.simp.rewrite] add_zero:1000:
+      0 + 0
+    ==>
+      0
```
b-mehta, line 48: This trace isn't consistent with the description you give above - if I'm traversing from the outside in, I see _ + 0 before I see Nat.fib 0, so your informal description would simplify first to `37 * Nat.fib 0`

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+⊢ 37 * (Nat.fib 0 + 0) = 0
+⊢ 37 * (0 + 0) = 0
+⊢ 37 * 0 = 0
+⊢ 0 = 0
+⊢ True
+```
+
+In this picture, simp lemmas are *fixed* rules to turn a *specific* left hand side into a *specific* right hand side.
+In contrast, simprocs are *flexible* rules to turn a *specific* left hand side into a right hand side *computed* from the left hand side.
+In this sense, they are *parametric simp lemmas*.
```
b-mehta, line 69: To a beginner, I don't think this is a useful distinction without some more explanation. The lemma `a + 0 = a` is "parametric" in the sense that I can feed in any `a`, and the lemma still applies, indeed `a` is a parameter here! There's also a reasonable interpretation of "specific" which makes this example not a simp lemma: it's flexible in what `a` can be, I haven't specified the value of `a`.

In particular, the description below, "combinatorial explosion of lemmas"  - to a beginner - doesn't clearly distinguish between `1 + 0 = 1`, `2 + 0 = 2`, `3 + 0 = 3`, etc as a combinatorial explosion of lemmas at this point either, I can imagine someone thinking `a + 0 = a` isn't solvable without simprocs from this explanation.

posts/simprocs-intro.md
b-mehta: Or perhaps after `simp only [existsAndEq]`

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+
+```lean
+example : ∃ (a : ℤ), a*a = 25 ∧ a = 5 := by
+  simp +arith only [existsAndEq, and_true]
+
+example : (∃ a, (∃ b, a + b = 5) ∧ a = 3) ↔ ∃ b, 3 + b = 5 := by
+  simp only [existsAndEq, and_true]
+```
+
+Roughly speaking, the way this metaprogram operates is a follows: whenever an expression of the form `∃ a, P a` with `P a = Q ∧ R`  is encountered:
```
b-mehta, line 90: Given that `existsAndEq` is changing at the moment, and might be replaced by something more complex, I'm not sure this is the best example...

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+example (T : Type) (P : Prop) (a b : T) : ite (P ∨ ¬ P) a b = a := by
+  -- Works since `simp` can simplify `P ∨ ¬ P` to `True`.
+  simp only [↓reduceIte]
+
+example : ite FermatLastTheorem 1 2 = 1 := by
+  --This fails because `simp` can't reduce `FermatLastTheorem` to `True`
+  simp only [↓reduceIte]
+```
+
+Internally, this simproc is a small metaprogram does the following whenever an expression of the form `ite P a b` is encountered:
```
b-mehta, line 117:
```suggestion
Internally, this simproc is a small metaprogram that does the following whenever an expression of the form `ite P a b` is encountered:
```

posts/simprocs-intro.md
b-mehta: Could be a nice compromise to just link to the documentation for `reduceIte` and suggest pressing "source" for the interested user.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+example : ¬ 2 ∣ 49 := by
+  simp only [Nat.reduceDvd, not_false_iff]
+
+example : a ∣ a * b := by
+  simp only [Nat.reduceDvd] --fails
+```
+
+Here the metaprogram run by `Nat.reduceDvd` does the following whenever an expression of the form `a ∣ b` (where `a, b` are natural numbers) is encountered:
+- Check that `∣` is the usual natural number division.
+- Try to extract explicit (literal) values for `a` and `b`.
```
b-mehta, line 171: At this point "literal" is pretty meaningless to the reader, and "extract explicit values" is ambiguous too - it could mean "compute a and b" as well: this line has a pretty high chance of being confusing.

posts/simprocs-intro.md
```diff
@@ -0,0 +1,550 @@
...
+Finally, we walk through an explicit example of a simproc for a simple custom function.
+
+## How `simp` works
+
+In this section we present some of the inner workings of `simp`.
+
+First we give an overview of the way `simp`, then we delve into the specifics by introducing:
+* The `SimpM` monad, which is the metaprogramming monad holding the information relevant to a `simp` call.
+* `Step`/`DStep`, the Lean representation of a single (definitional) simplification step.
+* `Simproc`/`DSimproc`, the Lean representation of (definitional) simprocs.
```
b-mehta, line 218:
```suggestion
* `Step` (`DStep`), the Lean representation of a single (definitional) simplification step.
* `Simproc` (`DSimproc`), the Lean representation of (definitional) simprocs.
```
or some other version - currently it's not clear what the correspondence actually is. You might also like to explicitly mention that the `D` is for definitional...

bryangingechen:
> That's definitely an option. What parts do you envision? The ones as we currently split them, or something like "How simp works" then "What are simprocs"?

I think the way you had it split into two parts would be a good starting point. Maybe there's even enough material for 3 or more posts? (How simp works, what are simprocs, how to write a simproc)

posts/simprocs-intro.md
Paul-Lez: That's a good point! I think it might be best to keep dsimprocs in the extras, since the discussion is almost identical and having both makes things a little cluttered imo.

posts/simprocs-intro.md
Paul-Lez: That's a good point! I guess this is a downside of being too informal in the first explanation of how simp works roughly. What I think is happening here is that lemmas are also treated as "post" simp procedures, i.e. simp will see `Nat.fib 0 + 0` before it sees `Nat.fib 0`, but it visits subexpressions of `Nat.fib 0 + 0` *before* trying to apply lemmas to `Nat.fib 0 + 0` (or rather what `Nat.fib 0 + 0` has been transformed to after its subexpressions have been visited). 

I'll try to reformulate this section to avoid giving the wrong impression.

posts/simprocs-intro.md
YaelDillies: You are right, they are a year and a half old by now. The PR that introduced them was https://github.com/leanprover/lean4/pull/3124

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,173 @@
...
+The second post will be a walkthrough to writing a simple simproc in three different ways.
+
+# What is a simproc
+
+To understand what a simproc is and how it works, we will first explain how `simp` works.
+Then we will give some examples and non-examples of simprocs as well as pointers to analogous concepts in other theorem provers.
+
+## How simp works
+
+On an informal level, `simp` can be seen as recursively traversing the expression `e` to be simplified from the outside-in, each time checking whether the expression matches the left hand side `LHS` of some simp lemma of the form `LHS = RHS`, and replacing `LHS` by `RHS` in the expression if it does match.
```
b-mehta, line 29: This appears inaccurate: note that the Lean reference manual says this: "The [simp](https://lean-lang.org/doc/reference/latest/Tactic-Proofs/Tactic-Reference/#simp) tactics primarily rewrite from the inside out."

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,173 @@
...
+## Examples of simprocs
+
+In this section, we exemplify three simprocs through the following use cases:
+* Computation
+* Avoiding combinatorial explosion of lemmas
+* Performance optimisation
+
+### Computation:
+
+#### The `Finset.Icc_ofNat_ofNAt` simproc
```
b-mehta:
```suggestion
#### The `Finset.Icc_ofNat_ofNat` simproc
```

posts/simprocs-intro.md
YaelDillies: To me, this is the best (and currently only?) example of something that clearly can't be done with finitely many lemmas. So I am willing to keep it even if the technical discussion part will become outdated.

Well in fact we've removed the technical discussion part ;)

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,173 @@
...
+
+The `Finset.Icc_ofNat_ofNAt` simproc is designed to take expressions of the form `Finset.Icc a b` where `a` and `b` are numerals, and simplify them to an explicit set.
```
b-mehta:
```suggestion
The `Finset.Icc_ofNat_ofNat` simproc is designed to take expressions of the form `Finset.Icc a b` where `a` and `b` are numerals, and simplify them to an explicit set.
```

posts/simprocs-intro.md
YaelDillies: All gone now

posts/simprocs-intro.md
YaelDillies: It's back, but with an explanation now

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,173 @@
...
+In this sense, they are *parametric simp lemmas*.
+
+## Examples of simprocs
+
+In this section, we exemplify three simprocs through the following use cases:
+* Computation
+* Avoiding combinatorial explosion of lemmas
+* Performance optimisation
+
+### Computation:
```
b-mehta, line 53: In each of the examples in this section, `decide` works to prove the goal, so it's not clear to me that this section works as a good example of why simprocs are useful.

posts/simprocs-intro.md
YaelDillies: I think this is **the** point of simprocs, so I really want to keep that distinction. The blurb now refers to the "shape" of the LHS and RHS, rather than to the LHS and RHS directly. Hopefully that clears things up enough?

posts/simprocs-intro.md
YaelDillies: The trace is gone but I think the underlying phrasing issue remains.

posts/how-to-write-simprocs.md
```diff
@@ -0,0 +1,377 @@
...
+
+First, we detail the inner workings of `simp` relevant to simprocs.
+Then we explain the syntax and general structure of a simproc.
+Finally, we walk through an explicit example of a simproc for a simple custom function.
+
+## How `simp` works
+
+In this section we present some of the inner workings of `simp`.
+
+First we give an overview of the way `simp`, then we delve into the specifics by introducing:
```
b-mehta, line 29:
```suggestion
First we give an overview of the way `simp` works, then we delve into the specifics by introducing:
```

posts/simprocs-intro.md
YaelDillies: I've made this line talk about numerals instead. I think that's plenty fine.

posts/how-to-write-simprocs.md
```diff
@@ -0,0 +1,377 @@
...
+  Procedures are by default postprocedures as oftentimes a procedure can only trigger after the inner expressions have been simplified.
+* *Preprocedures* are called on an expression `e` before subexpressions of `e` are simplified.
+  Preprocedures are mostly used when the simplification order induced by a postprocedure would otherwise be inefficient by visiting irrelevant subexpressions first.
+  Preprocedures are associated with the `↓` symbol in several syntaxes throughout the simp codebase.
+
+The general rule of thumb is that postprocedures simplify from the inside-out, while preprocedures simplify from the outside-in.
+
+Roughly speaking, when acting on an expression `e`, `simp` does a combination of the following:
+- Run preprocedures, which may change `e`, then attempt to simplify subexpressions of `e`.
+- Run postprocedures *after* subexpressions of `e` have been simplified.
```
b-mehta, line 58:
```suggestion
Roughly speaking, when acting on an expression `e`, `simp` does a combination of the following:
- Run preprocedures, which may change `e`.
- Attempt to simplify subexpressions of `e`.
- Run postprocedures *after* subexpressions of `e` have been simplified.
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,173 @@
...
+
+Internally, this simproc is a small metaprogram that does the following whenever an expression of the form `ite P a b` is encountered:
+- Call `simp` on `P` to get a simplified expression `P'` and a proof `h` that `P = P'`.
+- If `P'` is `True` then return the simplified expression `a` and the proof `ite_cond_eq_true r` that `ite P a b = a`
+- If `P'` is `False` then return the simplified expression `b` and the proof `ite_cond_eq_false r` that `ite P a b = b`
+- Otherwise, let `simp` continue the search.
+
+### Many more applications!
+
+At the end of this blog post, we will see how to build step by step a simproc for computing a variant of `List.range` when the parameter is a natural number literal.
```
b-mehta, line 165: I presume you want to change this :)

posts/fantastic-simprocs.md
YaelDillies: This is the same phrasing issue as here: https://github.com/leanprover-community/blog/pull/97#discussion_r2040673512

namely simp *traverses* from the outside-in, and *rewrites* from the inside-out

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,172 @@
...
+example : a ∣ a * b := by
+  simp only [Nat.reduceDvd] --fails
+```
+
+Here the metaprogram run by `Nat.reduceDvd` does the following whenever an expression of the form `a ∣ b` (where `a, b` are natural numbers) is encountered:
+- Check that `∣` is the usual natural number division.
+- Check that `a` and `b` are numerals.
+- Compute `b % a`.
+- If this quantity is zero, then return `True` together with the proof `Nat.dvd_eq_true_of_mod_eq_zero a b rfl`.
+- If this quantity isn't zero, then return `False` together with the proof `Nat.dvd_eq_false_of_mod_ne_zero a b rfl`.
```
YaelDillies, line 96: I am not sure how I feel about these technicalish explanations being so early in the blog post. Should they be relegated outside of `## Examples`? to the second blog post? I am open to suggestions

posts/fantastic-simprocs.md
b-mehta: Right, and I think that since the default and common case is postprocedures, it makes more sense for the informal description to talk about that case instead.

YaelDillies (commented):
@b-mehta and @kbuzzard, thank you both for the candid review! I believe this is ready for a second round of review. If you can do it before tomorrow afternoon, then Paul and I will get to think about the required changes together while I am in London.

Also please don't merge! Just approve. Then we will merge the first blog post, and open a PR for the second one, because
* it looks like comments have been mostly concentrated on the first part so far
* the second part is much more complicated technically and I would like Kyle's opinion
* I would like the second blog post to appear a week or two after the first

posts/fantastic-simprocs.md
b-mehta: I'm pretty content with this as it is: you mention the right hand side is computed from the left, and giving an overview of how that computation could work is helpful imo. I don't see this as too technical (although you probably could omit the part about checking `∣`)

posts/fantastic-simprocs.md
Paul-Lez: Maybe it's worth mentioning also that any lemma (and simproc) can be set to act during the outside-in part, by adding the `\d` arrow in front of the identifier.

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,172 @@
...
+The down arrow `↓` in `simp only [↓reduceIte]` indicates that `reduceIte` is being called as a *preprocedure*.
+This means that `reduceIte` is run *before* the expressions `a` and `b` are simplified.
+Concretely, `simp` needs only simplify one of the expressions `a` and `b`, as the other one is discarded upfront.
+For example, `↓reduceIte` allows `simp` to avoid touching `bigComplicatedExpression` at all in the following:
+```
+example : (if 37 * 0 = 0 then 0 else bigComplicatedExpression) = 0 := by
+  simp only [↓reduceIte]
+```
+
+Notice that one could always replace `simp [↓reduceIte]` with `simp [ite_cond_eq_true, ite_cond_eq_false]`, but the latter would simplify both branches of the `ite`, instead of merely the one that is actually relevant.
```
b-mehta, line 159: I misunderstood this at first read. What about:
```suggestion
Notice that one could always replace `simp [↓reduceIte]` with `simp [ite_cond_eq_true, ite_cond_eq_false]` which would leave the same final goal state, but the would simplify inside both branches of the `ite`, instead of simplifying away the irrelevant one before passing to subexpressions.
```

posts/fantastic-simprocs.md
Paul-Lez: I agree - this is something I meant to remove (e.g. "Check that `∣` is the usual natural number division." is a very confusing sentence if you can't see the corresponding code)

b-mehta, line 100: At this point in the story, it's not clear to the reader what's meant by "return". I think it would be helpful to give a rough intuition of how a simproc works, before the examples start, saying something like "the simproc's job is to try to find an appropriate RHS, and if so, return it together with a proof that LHS = RHS. Then `simp` will perform this replacement, using the returned proof term as appropriate." (Adjust phrasing to your taste!)

posts/fantastic-simprocs.md
Paul-Lez: I've also pushed some local changes that I had forgotten to add, which remove the "outside-in" from this sentence and give a tiny bit more information about how simp traverses expressions.

posts/simprocs-intro.md
Paul-Lez: I hadn't pushed some of the local changes:) - the phrasing has now changed.

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,176 @@
...
+
+Internally, this simproc is a small metaprogram that does the following whenever an expression of the form `ite P a b` is encountered:
+- Call `simp` on `P` to get a simplified expression `P'` and a proof `h` that `P = P'`.
+- If `P'` is `True` then return the simplified expression `a` and the proof `ite_cond_eq_true r` that `ite P a b = a`
+- If `P'` is `False` then return the simplified expression `b` and the proof `ite_cond_eq_false r` that `ite P a b = b`
+- Otherwise, let `simp` continue the search.
```
b-mehta, line 165:
```suggestion
```
This is not mentioned in any of the other simproc descriptions!

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,176 @@
...
+
+<!-- TEASER_END -->
+
+This post describes purely informally what simprocs are and do.
+The second post will be a walkthrough to writing a simple simproc in three different ways.
+
+# What is a simproc
+
+To understand what a simproc is and how it works, we will first explain how `simp` works.
+Then we will give some examples and non-examples of simprocs as well as pointers to analogous concepts in other theorem provers.
```
b-mehta, line 25: Perhaps I'm missing something, but I don't actually see an explanation of what simprocs do here! You explain what simp does, then have a single sentence giving one intuition for what a simproc is, but you don't say what it actually does. And then you jump to the examples. You also mention non-examples and pointers to other theorem provers, but I don't see that in this post either?

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,176 @@
...
+> To see what proof steps `simp` used, one can use `set_option trace.Meta.Tactic.simp true`, which prints some of this information in the infoview.
+
+In this picture, simp lemmas are *fixed* rules to turn a left hand side of a *specific* shape into a right hand side of a *specific* shape.
+In contrast, simprocs are *flexible* rules to turn a left hand side of a *specific* shape into a right hand side of a shape *computed* from the left hand side.
+In this sense, they are *parametric simp lemmas*.
+To make the previous description a little more precise, the way `simp` works is by recursively traversing an expression `e` (and its subexpressions), and trying to match these with the "left hand side" of a simproc.
+
+## Examples of simprocs
+
+In this section, we exemplify three simprocs through the following use cases:
```
b-mehta, line 52:
```suggestion
In this section, we exemplify four simprocs through the following use cases:
```

posts/fantastic-simprocs.md
YaelDillies: There are now several sentences. I'm not necessarily happy with the amount of explanations though

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,198 @@
...
+⊢ 37 * 0 = 0
+⊢ 0 = 0
+⊢ True
+```
+
+> The order in which `simp` traverses an expression is relatively intricate.
+> As a simple approximation, simplification is performed from the inside-out.
+> See the next blog post for full details on the algorithm.
+
+> If you write `set_option trace.Meta.Tactic.simp true in example : MyGoal := by simp`, you will see the list of simplification steps `simp` performs on `Goal`.
```
b-mehta, line 50: Do you definitely want all of this as two block quotes? I don't think they're significant enough for that!

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,198 @@
...
+Even though the right hand side of a lemma changes depending on its left hand side, it only does so via *syntactic substitution*.
+
+For example, the lemma `Nat.mul_add (a b c : Nat) : a * (b + c) = a * b + a * c` can be specialized to have right hand side `0 * b + 0 * c`, `a * b + a * (c + d)`, etc...
+But all these have *shape* `_ * _ + _ * _` and eg no value of `a`, `b`, `c` will make it have shape `_ * _ + _ * _ + _ * _`.
+
+With this in mind, simprocs are *parametric simp lemmas*.
+
+## Everything is a simproc
+
+As a very neat unification of concepts, it turns out that `simp` only ever handles simprocs.
```
b-mehta, line 70:
```suggestion
As a very neat unification of concepts, it turns out that the implementation of `simp` only ever handles simprocs.
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,198 @@
...
+# What is a simproc?
+
+Earlier, we said "almost all rewriting rules are lemmas".
+What about rewriting rules that are not lemmas?
+This is where simprocs enter the picture.
+
+A simproc is a program which, given an expression `LHS`, computes a new expression `RHS` and finds a proof of `LHS = RHS` on the fly.
+
+The concept of a simproc is genuinely more powerful than that of a simp lemma.
+Even though the right hand side of a lemma changes depending on its left hand side, it only does so via *syntactic substitution*.
```
b-mehta, line 61:
```suggestion
This is because even though the right hand side of a lemma changes depending on its left hand side, it only does so via *syntactic substitution*.
```
or some other phrasing which clarifies that this sentence is an explanation of the previous

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,198 @@
...
+example : Finset.Icc 1 1 = {1} := by
+  simp only [Icc_ofNat_ofNat]
+
+example : Finset.Icc 1 4 = {1, 2, 3, 4} := by
+  simp only [Icc_ofNat_ofNat]
+
+example : Finset.Icc a (a + 2) = {a, a + 1, a + 2} := by
+  -- fails: the bounds of the interval aren't numerals!
+  simp only [Icc_ofNat_ofNat]
+```
```
b-mehta, line 99: imo this one is less compelling than the `Nat.reduceDvd` one, and it currently doesn't have the informal description that all the others do. I don't mind keeping it, but I think it should go after the reduceDvd one, and have some explanation of the metaprogram (which can be one/two lines)

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,198 @@
...
+
+example : (if FermatLastTheorem then 1 else 2) = 1 := by
+  --This fails because `simp` can't simplify `FermatLastTheorem` to `True` or `False`
+  simp only [↓reduceIte]
+  -- Remaining goal: `⊢ (if FermatLastTheorem then 1 else 2) = 1`
+  sorry -- See https://imperialcollegelondon.github.io/FLT for how to solve this `sorry` ;)
+```
+
+The down arrow `↓` in `simp only [↓reduceIte]` indicates that `reduceIte` is being called as a *preprocedure*.
+This means that `reduceIte` is run *before* the expressions `a` and `b` are simplified.
```
b-mehta, line 174: You could mention here that this is an example of why "simplification is performed from the inside-out" was a simplification (no pun intended), which a) gives a tiny bit more context on what it actually happening and b) makes it clearer that you're not self-contradicting

posts/fantastic-simprocs.md
YaelDillies: Hmm, your suggestion sounds a bit clunky

b-mehta: The distinction I want to draw here is that you first said `simp` is made of two components, then you're saying that actually the first one was a lie. And so adding "implementation" here makes it clearer that these are consistent viewpoints, one is how simp intuitively works and the other is how it's practically implemented. I'm happy with alternate phrasing to a similar effect, but I think the current phrasing gives a confusing message here.

posts/fantastic-simprocs.md
YaelDillies: Oh I thought that was pretty nice. Imagining myself reading this blog post as a beginner, these block quotes are answering the exact two questions I think I would be asking myself.

Or do you mean that there's better formatting than a block quote? I was using that as an aparté

b-mehta: I think they can just be text as usual: block quotes in this way are usually used to indicate key takeaways from the overall post; these are useful points but I don't think they need highlighting in this way.

b-mehta: But this is predicated on the way block quotes are displayed in the blog; perhaps the disagreement here is just that we're picturing different things!

posts/fantastic-simprocs.md
YaelDillies: I've added an explanation of why `decide` bad. Hopefully that's enough

posts/simprocs-intro.md
YaelDillies: I've added a disclaimer

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Computation
+
+Computations are an integral part of theorem proving, and as such there are many ways to perform them.
+For example, you will find that the `decide` tactic closes all the examples in this subsection.
+There are a few reasons why simprocs are interesting for computation regardless:
+* **`decide` relies on decidability instances**.
+  Not everything one may want to compute is decidable, and not every decidability instance is efficient.
+  In fact, most `Decidable` instances in Lean and Mathlib are very generic, and therefore unspecific and inefficient.
+  Using a simproc gives the opportunity to use a domain-specific algorithm, which is more likely to be efficient and does not rely the decidability machinery.
```
b-mehta, line 114:
```suggestion
  Using a simproc gives the opportunity to use a domain-specific algorithm, which is more likely to be efficient and does not rely on the decidability machinery.
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,218 @@
...
+has_math: true
+link: ''
+slug: fantastic-simprocs
+tags: 'simp, simproc, meta'
+title: Fantastic Simprocs and Where to Find Them.
+type: text
+---
+
+Lean core added support for custom simplification procedures, aka *simprocs*.
+This blog post is the first in a series of two aimed at explaining what a simproc is, what kind of problems can be solved with simprocs, and what tools we have to write them.
```
eric-wieser, line 15: Style question; should we write `` `simproc` `` everywhere?

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,218 @@
...
+> As a simple approximation, simplification is performed from the inside-out.
+> See the next blog post for full details on the algorithm.
+
+> If you write `set_option trace.Meta.Tactic.simp true in example : MyGoal := by simp`, you will see the list of simplification steps `simp` performs on `Goal`.
+
+# What is a simproc?
+
+Earlier, we said "almost all rewriting rules are lemmas".
+What about rewriting rules that are not lemmas?
+This is where simprocs enter the picture.
```
eric-wieser, line 56: Where do theorems like `: Even 0` fit into this narrative? Perhaps worth a footnote that theorem of the form `P` are often treated as `P ↔ True`?

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,218 @@
...
+
+To understand what a simproc is and how it works, we will first explain how `simp` works.
+Then we will give some examples and non-examples of simprocs.
+
+# How simp works
+
+`simp` is made of two components.
+
+The first component is **rewriting rules**.
+Almost all rewriting rules are lemmas that prove an equality and are tagged with `@[simp]` in eg Lean or Mathlib.
```
eric-wieser, line 30:
```suggestion
Almost all rewriting rules are lemmas that prove an equality or `↔` and are tagged with `@[simp]` in eg Lean or Mathlib.
```

posts/fantastic-simprocs.md
YaelDillies: ... or `Simproc`? I have been following the convention that formal concepts are quoted, while informal ones aren't. In this blog post, a simproc is an informal concept (but `simp` is a formal one), and simprocs become the formal `Simproc`s in the followup blog post

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,218 @@
...
+example (p : Nat → Prop) : ∃ x : Nat, p x ∧ x = 5 := by
+  simp only [existsAndEq]
+  -- Remaining goal: `⊢ p 5`
+
+example (p q : Nat → Prop) : ∃ x : Nat, p x ∧ x = 5 ∧ q x := by
+  simp only [existsAndEq]
+  -- Remaining goal: `⊢ p 5 ∧ q 5`
+```
+
+When presented with a left hand side of the form `∃ x, P x, where `P x` is of the form `_ ∧ ... ∧ _`, `existsAndEq` does the following:
```
eric-wieser, line 94:
```suggestion
When presented with a left hand side of the form `∃ x, P x`, where `P x` is of the form `_ ∧ ... ∧ _`, `existsAndEq` does the following:
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,218 @@
...
+- Recursively traverse `P x` inside the existential quantifier looking for an equality `x = a` for some `a`.
+- If an equality is found, construct a proof that `∀ x, p x → x = a`.
+- Return the right hand side `P a` together with the proof obtained from the following lemma:
+```lean
+lemma exists_of_imp_eq {α : Sort u} {p : α → Prop} (a : α) (h : ∀ x, p x → x = a) :
+    (∃ x, p x) = p a
+```
```
eric-wieser, line 101:
````suggestion
- Return the right hand side `P a` together with the proof obtained from the following lemma:
  ```lean
  lemma exists_of_imp_eq {α : Sort u} {p : α → Prop} (a : α) (h : ∀ x, p x → x = a) :
      (∃ x, p x) = p a
  ```
````

File: posts/fantastic-simprocs.md [hunk 45]
```diff
@@ -0,0 +1,218 @@
...
+## Many more applications!
+
+In the second blog post, we will see how to build step by step a simproc for computing a variant of `List.range` when the parameter is a numeral.
+
+# A few caveats
+
+The current design of simprocs comes with a few restrictions that are worth keeping in mind:
+* By definition, **a simproc can only be used in `simp`** (and `simp`-like tactics like `simp_rw`, `simpa`, `aesop`), even though the notion of a "modular lemma" could be useful in other rewriting tactics like `rw`.
+* **One cannot provide arguments to a simproc to restrict the occurrences it rewrites**.
+  In contrast, this is possible for lemmas in all rewriting tactics: eg `rw [add_comm c]` turns `⊢ a + b = c + d` into `⊢ a + b = d + c` where `rw [add_comm]` would instead have turned it into `⊢ b + a = c + d`.
```
eric-wieser, line 218: I was hoping this would work, but alas:
```lean
def myProc (a : Nat) : Lean.Meta.Simp.Simproc := fun e => do
  return .continue

-- illegal
simproc_pattern% _ => myProc 

-- ok, but arguments were consumed already
def wat := myProc 1
simproc_pattern% _ => wat
```

posts/fantastic-simprocs.md
eric-wieser: Both `Lean.Meta.Simp.Simproc` and `simproc` are valid pieces of lean syntax, so I think the latter is fine to use instead of the former.

My heuristic is "if it's not a real word then always write and quote it as code", but it's not my blog post so I'll let you make the final call.

b-mehta: For comparison, the reference manual also refers to simproc without code indicators, so this is at least consistent with that

posts/fantastic-simprocs.md
YaelDillies: I've added an explanation earlier

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,218 @@
...
+description: 'Informal introduction to simprocs and what they are useful for'
+has_math: true
+link: ''
+slug: fantastic-simprocs
+tags: 'simp, simproc, meta'
+title: Fantastic Simprocs and Where to Find Them.
+type: text
+---
+
+Lean core added support for custom simplification procedures, aka *simprocs*.
```
eric-wieser, line 14:
```suggestion
Lean v4.6.0 added support for custom simplification procedures, aka *simprocs*.
```
perhaps with some remark about this being a long time ago!

posts/fantastic-simprocs.md
eric-wieser: Can we link the reference manual from the blog post?

b-mehta: Sure, I think it makes sense to mention it at the end of the "how simp works" section, saying something like, you can read about it in more detail here: https://lean-lang.org/doc/reference/latest////The-Simplifier/#the-simplifier

posts/fantastic-simprocs.md
YaelDillies: Yes sadly I think a few things to change for Lean to acquire `rwproc`

posts/fantastic-simprocs.md
b-mehta: There's not much about simprocs there, only a paragraph here https://lean-lang.org/doc/reference/latest////The-Simplifier/Rewrite-Rules/#simp-rewrites

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,219 @@
...
+⊢ 37 * 0 = 0
+⊢ 0 = 0
+⊢ True
+```
+
+> The order in which `simp` traverses an expression is relatively intricate.
+> As a simple approximation, simplification is performed from the inside-out.
+> See the next blog post for full details on the algorithm.
+
+> If you write `set_option trace.Meta.Tactic.simp true in example : MyGoal := by simp`, you will see the list of simplification steps `simp` performs on `Goal`.
```
adomani, line 51:
```suggestion
> If you write `set_option trace.Meta.Tactic.simp true in example : MyGoal := by simp`, you will see the list of simplification steps `simp` performs on `MyGoal`.
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,219 @@
...
+⊢ 37 * (Nat.fib 0 + 0) = 0
+⊢ 37 * (0 + 0) = 0
+⊢ 37 * 0 = 0
+⊢ 0 = 0
+⊢ True
+```
+
+> The order in which `simp` traverses an expression is relatively intricate.
+> As a simple approximation, simplification is performed from the inside-out.
+> See the next blog post for full details on the algorithm.
```
adomani, line 49:
```suggestion
> See the next blog post for more details on the algorithm.
```
unless you really mean *full*!

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,219 @@
...
+
+When presented with a left hand side of the form `a ∣ b` where `a` and `b` are natural numbers, `Nat.reduceDvd` does the following:
+- Check that `a` and `b` are numerals.
+- Compute `b % a`.
+- If `b % a` is zero, then return the right hand side `True` together with the proof `Nat.dvd_eq_true_of_mod_eq_zero a b rfl`.
+- If `b % a` isn't zero, then return the right hand side `False` together with the proof `Nat.dvd_eq_false_of_mod_ne_zero a b rfl`.
+
+### The `Finset.Icc_ofNat_ofNat` simproc
+
+The `Finset.Icc_ofNat_ofNat` simproc is designed to take expressions of the form `Finset.Icc a b` where `a` and `b` are numerals, and simplify them to an explicit set.
```
b-mehta, line 146: It might be worth giving a link to docs for `Finset.Icc`; I'd imagine some readers might not know what it is!

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,219 @@
...
+
+example : a ∣ a * b := by
+  simp only [Nat.reduceDvd] --fails
+```
+
+When presented with a left hand side of the form `a ∣ b` where `a` and `b` are natural numbers, `Nat.reduceDvd` does the following:
+- Check that `a` and `b` are numerals.
+- Compute `b % a`.
+- If `b % a` is zero, then return the right hand side `True` together with the proof `Nat.dvd_eq_true_of_mod_eq_zero a b rfl`.
+- If `b % a` isn't zero, then return the right hand side `False` together with the proof `Nat.dvd_eq_false_of_mod_ne_zero a b rfl`.
```
b-mehta, line 142: In the ite example you give the proof as well as saying what it's a proof of, I think it's clearer to do the same here too:
```suggestion
- If `b % a` isn't zero, then return the right hand side `False` together with the proof `Nat.dvd_eq_false_of_mod_ne_zero a b rfl` that `b % a = 0 = False`.
```

b-mehta: (and similarly for the `True` case)

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,219 @@
...
+  simp only [Icc_ofNat_ofNat]
+```
+
+When presented with a left hand side of the form `Finset.Icc a b` where `a` and `b` are natural numbers, `Finset.Icc_ofNat_ofNat` does the following:
+* Check that `a` and `b` are numerals.
+* compute the expression `{a, ..., b}` recursively on `b`, along with the proof that it equals `Finset.Icc a b`.
+
+## Performance optimisation: The `reduceIte` simproc
+
+The `reduceIte` simproc is designed to take expressions of the form `if P then a else b` (aka `ite P a b`) and replace them with `a` or `b`, depending on whether `P` simplify to `True` or `False`.
```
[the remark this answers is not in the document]
YaelDillies: [The Cambridge dictionary lists "aka"](https://dictionary.cambridge.org/dictionary/english/aka#google_vignette)

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+Computations are an integral part of theorem proving, and as such there are many ways to perform them.
+For example, you will find that the `decide` tactic closes all the examples in this subsection.
+There are a few reasons why simprocs are interesting for computation regardless:
+* **`decide` relies on decidability instances**.
+  Not everything one may want to compute is decidable, and not every decidability instance is efficient.
+  In fact, most `Decidable` instances in Lean and Mathlib are very generic, and therefore unspecific and inefficient.
+  Using a simproc gives the opportunity to use a domain-specific algorithm, which is more likely to be efficient and does not rely on the decidability machinery.
+* **`decide` cannot compute the right hand side**, given the left hand side only.
+  `decide` only works on goals that do not contain any metavariable.
+  This rules out using `decide` to discover out what a left hand side is equal to.
```
adomani, line 118:
```suggestion
  This rules out using `decide` to discover what a left hand side is equal to.
```
or maybe "find out"?

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+When presented with a left hand side of the form `Finset.Icc a b` where `a` and `b` are natural numbers, `Finset.Icc_ofNat_ofNat` does the following:
+* Check that `a` and `b` are numerals.
+* compute the expression `{a, ..., b}` recursively on `b`, along with the proof that it equals `Finset.Icc a b`.
+
+## Performance optimisation: The `reduceIte` simproc
+
+The `reduceIte` simproc is designed to take expressions of the form `if P then a else b` (aka `ite P a b`) and replace them with `a` or `b`, depending on whether `P` simplify to `True` or `False`.
+
+```lean
+example : (if 37 * 0 then 1 else 2) = 1 := by
```
adomani, line 174:
```suggestion
example : (if 37 * 0 = 0 then 1 else 2) = 1 := by
```
this is missing right?  :smile:

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+  -- Works since `simp` can simplify `37 * 0` to `True`
```
adomani, line 175:
```suggestion
  -- Works since `simp` can simplify `37 * 0 = 0` to `True`
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+> This is a concrete example where simplification happens outside-in.
+
+Concretely, `simp` needs only simplify one of the expressions `a` and `b`, as the other one is discarded upfront.
+For example, `↓reduceIte` allows `simp` to avoid touching `bigComplicatedExpression` at all in the following:
+```
+example : (if 37 * 0 = 0 then 0 else bigComplicatedExpression) = 0 := by
+  simp only [↓reduceIte]
+```
+
+Notice that `simp [↓reduceIte]` is equivalent to `simp [ite_cond_eq_true, ite_cond_eq_false]`, but the latter simplifies inside both branches of the `ite`, instead of simplifying away the irrelevant one before passing to subexpressions.
```
adomani, line 205:
```suggestion
Notice that `simp [↓reduceIte]` is ultimately equivalent to `simp [ite_cond_eq_true, ite_cond_eq_false]`, but the latter simplifies inside both branches of the `ite`, instead of simplifying away the irrelevant one before passing to subexpressions.
```
since otherwise "equivalent" could be interpreted as "behaves the same", but then you explain that it does not!

posts/fantastic-simprocs.md
b-mehta: You could also link https://leanprover-community.github.io/extras/simp.html#introduction

digama0: My vote is to italicize it on first use but otherwise treat it like a regular word (which seems to be already happening).

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
+---
+author: 'Yaël Dillies, Paul Lezeau'
+category: 'Metaprogramming'
+date: 2025-04-16 12:00:00 UTC+01:00
+description: 'Informal introduction to simprocs and what they are useful for'
+has_math: true
+link: ''
+slug: fantastic-simprocs
+tags: 'simp, simproc, meta'
+title: Fantastic Simprocs and Where to Find Them.
```
tb65536, line 10: I know the title is a slight joke, but I would actually be curious to know if there is a way to find or search for simprocs in mathlib? E.g., if I'm in the middle of a proof, is there some way for me to check if there is a simproc that would be helpful here? (Right now it seems to me like I have to keep track of what simprocs mathlib has and what their names are).

eric-wieser:
```lean
run_meta do
  let s ← Lean.Meta.Simp.getSimprocs
  Lean.logInfo m!"{s.pre.toArray ++ s.post.toArray}"
  pure ()
```

eric-wieser: I think that only finds `simproc`s (the ones enabled automatically), not `simproc_decls` (the ones you refer to by name). Loogle should be able to find both.

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+
+example : a ∣ a * b := by
+  simp only [Nat.reduceDvd] --fails
+```
+
+When presented with a left hand side of the form `a ∣ b` where `a` and `b` are natural numbers, `Nat.reduceDvd` does the following:
+- Check that `a` and `b` are numerals.
+- Compute `b % a`.
+- If `b % a` is zero, then return the right hand side `True` together with the proof `Nat.dvd_eq_true_of_mod_eq_zero a b rfl` that `b % a = 0 = True`.
+- If `b % a` isn't zero, then return the right hand side `False` together with the proof `Nat.dvd_eq_false_of_mod_ne_zero a b rfl` that `b % a = 0 = False`.
```
erdOne, line 142:
```suggestion
- If `b % a` is zero, then return the right hand side `True` together with the proof `Nat.dvd_eq_true_of_mod_eq_zero a b rfl` that `(b % a = 0) = True`.
- If `b % a` isn't zero, then return the right hand side `False` together with the proof `Nat.dvd_eq_false_of_mod_ne_zero a b rfl` that `(b % a = 0) = False`.
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+
+### The `Finset.Icc_ofNat_ofNat` simproc
+
+`Finset.Icc a b` for `a` and `b` in a partial order is the finite set of elements lying between `a` and `b`.
+
+The `Finset.Icc_ofNat_ofNat` simproc is designed to take expressions of the form [`Finset.Icc a b`](https://leanprover-community.github.io/mathlib4_docs/search.html?q=Finset.Icc) where `a` and `b` are numerals, and simplify them to an explicit set.
```
erdOne, line 148: Should this not link to the actual docs page instead? Or is that being less stable a problem?

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+  simp only [Icc_ofNat_ofNat]
+```
+
+When presented with a left hand side of the form `Finset.Icc a b` where `a` and `b` are natural numbers, `Finset.Icc_ofNat_ofNat` does the following:
+* Check that `a` and `b` are numerals.
+* compute the expression `{a, ..., b}` recursively on `b`, along with the proof that it equals `Finset.Icc a b`.
+
+## Performance optimisation: The `reduceIte` simproc
+
+The `reduceIte` simproc is designed to take expressions of the form `if P then a else b` (aka `ite P a b`) and replace them with `a` or `b`, depending on whether `P` simplify to `True` or `False`.
```
erdOne, line 171:
```suggestion
The `reduceIte` simproc is designed to take expressions of the form `if P then a else b` (aka `ite P a b`) and replace them with `a` or `b`, depending on whether `P` simplifies to `True` or `False`.
```

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+## Many more applications!
+
+In the second blog post, we will see how to build step by step a simproc for computing a variant of `List.range` when the parameter is a numeral.
+
+# A few caveats
+
+The current design of simprocs comes with a few restrictions that are worth keeping in mind:
+* By definition, **a simproc can only be used in `simp`** (and `simp`-like tactics like `simp_rw`, `simpa`, `aesop`), even though the notion of a "modular lemma" could be useful in other rewriting tactics like `rw`.
+* **One cannot provide arguments to a simproc to restrict the occurrences it rewrites**.
+  In contrast, this is possible for lemmas in all rewriting tactics: e.g. `rw [add_comm c]` turns `⊢ a + b = c + d` into `⊢ a + b = d + c` where `rw [add_comm]` would instead have turned it into `⊢ b + a = c + d`.
```
erdOne, line 221: Is this true? Can you not do `P : Nat -> Simproc` and `simp [P 37]`?

posts/fantastic-simprocs.md
eric-wieser: This should link to the magic links that Zulip uses.

posts/fantastic-simprocs.md
eric-wieser: [no](https://github.com/leanprover-community/blog/pull/97#discussion_r2045348464)

YaelDillies: As explained in the other blog post (#98), simp doesn't consume bare elements of type `Simproc`, but instead elements of type `Simproc` annotated with extra data, like whether it's pre or post. So `simp [(something : Simproc)]` doesn't even make sense

YaelDillies: That too. "It wasn't intended" is probably the answer

posts/fantastic-simprocs.md
Paul-Lez: It would be pretty cool to have a command ``#simprocs` that takes in a pattern and outputs the list of `simproc`s that activate on that pattern, e.g something like `#simprocs (ite _ _ _)` would find `iteReduce`

posts/fantastic-simprocs.md
b-mehta: I think it's a bit more than just the parser: there's also a check to make sure what you passed in was `Simproc` (and not, eg `_ -> Simproc`), and the simprocs are stored in an array which is also expecting stuff of that type. So "it wasn't intended" sounds closer to the truth

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+example : (if 37 * 0 = 0 then 0 else bigComplicatedExpression) = 0 := by
+  simp only [↓reduceIte]
+```
+
+Notice that `simp [↓reduceIte]` is ultimately equivalent to `simp [ite_cond_eq_true, ite_cond_eq_false]`, but the latter simplifies inside both branches of the `ite`, instead of simplifying away the irrelevant one before passing to subexpressions.
+
+When presented with a left hand side of the form `ite P a b`, `reduceIte` does the following:
+- Call `simp` on `P` to get a simplified expression `P'` and a proof `h` that `P = P'`.
+- If `P'` is `True` then return the right hand side `a` together with the proof `ite_cond_eq_true r` that `ite P a b = a`.
+- If `P'` is `False` then return the right hand side `b` together with the proof `ite_cond_eq_false r` that `ite P a b = b`.
```
kbuzzard, line 210: I have now read and understood all of this example and am completely confused about why we need a `simproc` to do this. If the *only* reason is "saves us from simplifying b if P is true and b is complicated" then I think this example should either be deleted, or rewritten to explain that this is the point. I am not entirely clear on why `simp` can't solve this without a simproc so my comments on this example should be taken with a pinch of salt, but if I'm right and the sole point is avoiding simplifying a complicated expression b then I think this example should *start* with: "Consider the following goal:"
```
example : (if 37 * 0 = 0 then 0 else bigComplicatedExpression) = 0 := by
```
and then explain the problem with running `simp`, and then explain the simproc, and then explain the down-arrow, and then write
```
  simp only [↓reduceIte]
```
and now everything is clear. 

Indeed this works fine:
```
example : 0 = if True then 0 else 1 := by
  simp [-reduceIte]
```
so it seems to me that the only point of this simproc is to save you from simplifying the irrelevant possible answer. You should say this right at the start.

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Many more applications!
+
+In the second blog post, we will see how to build step by step a simproc for computing a variant of [`List.range`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=List.range#doc) when the parameter is a numeral.
+
+# A few caveats
+
+The current design of simprocs comes with a few restrictions that are worth keeping in mind:
+* By definition, **a simproc can only be used in `simp`** (and `simp`-like tactics like `simp_rw`, `simpa`, `aesop`), even though the notion of a "modular lemma" could be useful in other rewriting tactics like `rw`.
```
kbuzzard, line 219: Why mention `aesop`? Why not also mention `norm_num`, that runs `simp`. You could say "`simp`** (and tactics which call `simp` such as `simpa`, `aesop` etc)

kbuzzard (commented):
This is looking great! I've left a host of minor comments which will hopefully make it even better.

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+  simp only [↓reduceIte]
+  -- Remaining goal: `⊢ (if FermatLastTheorem then 1 else 2) = 1`
+  sorry -- See https://imperialcollegelondon.github.io/FLT for how to solve this `sorry` ;)
+```
+
+The down arrow `↓` in `simp only [↓reduceIte]` indicates that `reduceIte` is being called as a *preprocedure*.
+This means that `reduceIte` is run *before* the expressions `a` and `b` are simplified.
+
+> Recall that we said that, as a simple approximation, simplification is performed from the inside-out.
+> This is a concrete example where simplification happens outside-in.
```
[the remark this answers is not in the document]
YaelDillies: We're simplifying `ite P a b` before `a` and `b`. This is what's "outside-in". It is true that in order to simplify `ite P a b` we also simplify `P`, but that's not what this remark is about.

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+
+In this section, we exemplify four simprocs through the following use cases:
+* Avoiding combinatorial explosion of lemmas
+* Computation
+* Performance optimisation
+
+## Avoiding combinatorial explosion of lemmas: The `existsAndEq` simproc
+
+The [`existsAndEq`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=existsAndEq#doc) simproc is designed to simplify expressions of the form `∃ x, ... ∧ x = a ∧ ...` where `a` is some quantity independent of `x` by removing the existential quantifier and replacing all occurences of `x` by `a`.
+
```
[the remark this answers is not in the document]
YaelDillies: Because it might move! The `find` page is what the Zulip linkifier uses, and you don't see people complaining that links in old posts are dead, so this is most definitely the right move.

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+  In fact, most `Decidable` instances in Lean and Mathlib are very generic, and therefore unspecific and inefficient.
+  Using a simproc gives the opportunity to use a domain-specific algorithm, which is more likely to be efficient and does not rely on the decidability machinery.
+* **`decide` cannot compute the right hand side**, given the left hand side only.
+  `decide` only works on goals that do not contain any metavariable.
+  This rules out using `decide` to find out what a left hand side is equal to.
+  One would need to write down the right hand side we are looking for in order for `decide` to show that it's equal to the left hand side.
+
+### The `Nat.reduceDvd` simproc
+
+The [`Nat.reduceDvd`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Nat.reduceDvd#doc) simproc is designed to take expressions of the form `a | b` where `a, b` are natural number literals, and simplify them to `True` or `False`.
```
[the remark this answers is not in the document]
YaelDillies: Thanks, I thought I had already replaced "literal" with "numeral" everywhere but apparently not

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+  simp only [Nat.reduceDvd]
+  -- Remaining goal: `¬ False`
+  simp
+
+example : a ∣ a * b := by
+  simp only [Nat.reduceDvd] --fails
+```
+
+When presented with a left hand side of the form `a ∣ b` where `a` and `b` are natural numbers, `Nat.reduceDvd` does the following:
+- Check that `a` and `b` are numerals.
```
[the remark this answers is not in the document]
YaelDillies: Indeed :)

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+example : (if 37 * 0 = 0 then 1 else 2) = 1 := by
+  -- Works since `simp` can simplify `37 * 0 = 0` to `True`
+  -- because it knows the lemma `mul_zero a : a * 0 = 0`.
+  simp only [↓reduceIte]
+
+example (X : Type) (P : Prop) (a b : X) : (if P ∨ ¬ P then a else b) = a := by
+  -- Works since `simp` can simplify `P ∨ ¬ P` to `True`.
+  simp only [↓reduceIte]
+
+open Classical
```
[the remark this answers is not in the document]
YaelDillies: We can remove it once you are done proving `FermatLastTheorem` 😉

posts/fantastic-simprocs.md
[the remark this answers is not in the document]
YaelDillies: I've clarified, hopefully it's better now

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+This is where simprocs enter the picture.
+
+A simproc is a program which, given an expression `LHS`, computes a new expression `RHS` and finds a proof of `LHS = RHS` on the fly.
+
+The concept of a simproc is genuinely more powerful than that of a simp lemma.
+This is because the right hand side of a lemma changes only via *syntactic substitution*.
+
+For example, the lemma `Nat.mul_add (a b c : Nat) : a * (b + c) = a * b + a * c` can be specialized to have right hand side `0 * b + 0 * c`, `a * b + a * (c + d)`, etc...
+But all these have *shape* `_ * _ + _ * _` and e.g. no value of `a`, `b`, `c` will make it have shape `_ * _ + _ * _ + _ * _`.
+
```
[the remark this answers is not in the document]
YaelDillies: The point I'm trying to make here is the distinction that Bhavik was explaining here: https://github.com/leanprover-community/blog/pull/97#discussion_r2021952082

posts/fantastic-simprocs.md
```diff
@@ -0,0 +1,221 @@
...
+Then we will give some examples and non-examples of simprocs.
+
+# How `simp` works
+
+`simp` is made of two components.
+
+The first component is **rewriting rules**.
+Almost all rewriting rules are lemmas that prove an equality `=` or iff `↔` and are tagged with `@[simp]` in e.g. Lean or Mathlib.
+A rewriting rule is characterised by its *left hand side* and *right hand side*.
+Eg for a lemma of the form `LHS = RHS` or `LHS ↔ RHS`, this is `LHS` and `RHS` respectively.
```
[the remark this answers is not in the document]
YaelDillies: If it reads patronising, then you haven't understood it! The point is that `LHS = RHS` is a very special case of a rewriting rule, and the general case is a simproc.

I don't think I can remove the explanation entirely either since I refer to the "left hand side" and "right hand side" of a rewriting rule several times below, in a precise sense.

I've tried something, but I'm not sure it's an improvement

posts/fantastic-simprocs.md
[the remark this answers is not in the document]
b-mehta: My reading of Kevin's message is that the point isn't clearly made in this paragraph: if a reader is confused by this they're not going to come to this PR and go and find my comment!

posts/fantastic-simprocs.md
Paul-Lez: I've changed the title to _Simprocs for the Working (Formal) Mathematician_, which I think make the more practical purpose of this post clearer (and we can repurpose the previous name for the last post in the series where we actually explain how to write a simproc)

posts/fantastic-simprocs.md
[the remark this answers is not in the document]
Paul-Lez: I think this paragraph has now been removed!

posts/fantastic-simprocs.md
```diff
@@ -118,6 +118,10 @@ The [`Nat.reduceDvd`](https://leanprover-community.github.io/mathlib4_docs/find/
 example : 3 ∣ 9 := by
   simp only [Nat.reduceDvd]
 
+-- `decide` cannot close this goal.
+example (f : Prop → Prop) : f (2 ∣ 4) = f True := by
```
b-mehta, line 122: This is a nice example!

posts/fantastic-simprocs.md
```diff
@@ -60,7 +60,7 @@ Indeed, we will soon see an example of a simproc taking the place of infinitely
 
 # Examples of simprocs
 
-In this section, we exemplify four simprocs through the following use cases:
+In this section, we exemplify four simprocs that cover the following following use cases:
```
eric-wieser, line 63: I'm not following the following following following. I think it should be removed.

PR #98, opened 2025-04-15
feat: Simp, made simple

YaelDillies:
This is the second part of #97.

Co-authored-by: Paul Lezeau <paul.lezeau@gmail.com>

### DIFF
--- a/posts/simp-made-simple.md
+++ b/posts/simp-made-simple.md
@@ -166,7 +166,7 @@ The constructors do the following:
 - `done` indicates that `simp` is done with a given expression.
   When `Nat.reduceDvd` is called on an expression of the form `a | b` where `a`, `b` are explicit natural numbers, it simplifies it to `True` or `False`. 
   Either way, the output is in simp normal form and there is no need to simplify it further.
-  Thus `Nat.reducedDvd` uses `done` in such a case.
+  Thus `Nat.reduceDvd` uses `done` in such a case.
 - `visit` indicates (for a pre-procedure) that a simplification has been done but that pre-procedures should be tried again on the simplified expression.
   When `reduceIte` is called on a expressions of the form `if p then a else b` where `p` can simplified to `True` (resp. `False`), it outputs `a` (resp. `b`). 
   Since `a` and `b` could be arbitrarily complicated expressions, it makes sense to try and simplify them further.
@@ -217,7 +217,7 @@ Let's go through these steps one by one.
   while we will often want to change the state during the `simp` call, we will rarely need to change the context.
   In programmer lingo, the context should be _immutable_.
   Thus, we use a different monad transformer called `ReaderT`, which is almost identical to `StateT`, but outputs a new monad where one can only read the type passed as parameter. 
-    > For completness: when working with `ReaderT`, one can still locally override the
+    > For completeness: when working with `ReaderT`, one can still locally override the
     > value of the variable that the monad keeps track of by using `withReader`. Intuitively, 
     > the difference between `State(Ref)T` and `ReaderT` is the following: 
     > - In `State(Ref)T`, one has access to a global variable that can be modified at will,
@@ -226,7 +226,7 @@ Let's go through these steps one by one.
 
 4) The final monad transformer application: `ReaderT Simp.MethodsRef $ ReaderT Simp.Context $ StateRefT Simp.State MetaM`. 
   This outputs a monad that has access to `Simp.Method` (passed via a ref). 
-  This capture the "pre" and "post" procedures that a given `simp` call can use, as well as the discharger that a given `simp` call can use, etc.
+  This captures the "pre" and "post" procedures that a given `simp` call can use, as well as the discharger that a given `simp` call can use, etc.
 
 ## `Simproc`s
 


posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+Often, when applying a theorem, we may need to provide additional proof terms for the hypotheses of the result. One useful feature of
+`simprocs` is that we can also call the discharger tactic provided to simp. Which discharger was provided by the user is part of the state stored by the `SimpM` monad, and can be access by the user via `Simp.Methods` (roughly speaking, the part of the state that encodes which methods `simp` can use to simplify an expression). `Simp.Methods` implements a function `discharge? : Expr → Option (Expr)` such that `discharge? goal` is equal to `some pf` if the discharger found a proof `pf` of `goal`, and none otherwise. Finally, to access the current "state" of `Simp.Methods`, one can use `Simp.getMethods`.
+
+In the following example, we implement a simproc that simplifies expressions of the form `(a * b).factorization ` to `a.factorization + b.factorization` whenever a proof that `a` and `b` are both non-zero can be found by the discharger.
+
+```lean
+open Qq
+
+simproc_decl factorizationMul (Nat.factorization (_ * _)) := fun e => do
+  let ⟨1, ~q(ℕ →₀ ℕ), ~q(Nat.factorization ($a * $b))⟩ ← inferTypeQ e | return .continue
```
eric-wieser, line 339: If we're using Qq here anyway, let's use `ofQ` to avoid needing to do the `inferTypeQ`.

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+To compute `revRange` using the dsimproc approach, we can do the following:
+```lean
+dsimproc_decl revRangeCompute (revRange _) := fun e => do
+  --Extract the natural number from the expression
+  let_expr revRange m ← e | return .continue
+  --Recover the natural number as a term of type `Nat`
+  let some n ← Nat.fromExpr? m | return .continue
+  let l := revRange n
+  --Convert the list to an `Expr`
+  return .visit <| ToExpr.toExpr l
```
eric-wieser, line 236: I think
```suggestion
  return .visit <| Lean.toExpr l
```
is the canonical spelling.

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+  let ⟨l, pf⟩ := go n
+  return .visit { expr := l, proof? := pf }
+```
+
+**Pros**:
+* Works regardless of definitional equalities.
+
+**Cons**:
+* Might involve a fair bit of meta code, a lot of which could *feel* like evaluating the function.
+
```
b-mehta, line 280:
```suggestion
**Cons**:
* Might involve a fair bit of meta code, a lot of which could *feel* like evaluating the function.
* Simplifying `revRange n` for a big input numeral `n` might produce a large proof term. In this specific case, the size of the produced proof term will be linear in `n`.
```

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+  | zero => aesop
+  | succ n h => rw [←hl] ; rfl
+
+open Qq in
+simproc_decl revRangeComputeProp (revRange _) := fun e => do
+  let_expr revRange m ← e | return .continue
+  let some n ← Nat.fromExpr? m | return .continue
+  let rec go (n : ℕ) : (l : Q(List ℕ)) × Q(revRange $n = $l) :=
+    match n with
+    | 0 => ⟨q(([] : List ℕ)), q(rfl)⟩
```
b-mehta, line 267: You are using a definitional equality here, that `revRange 0` is defined as the empty list!

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+
+To compute `revRange` using the dsimproc approach, we can do the following:
+```lean
+dsimproc_decl revRangeCompute (revRange _) := fun e => do
+  --Extract the natural number from the expression
+  let_expr revRange m ← e | return .continue
+  --Recover the natural number as a term of type `Nat`
+  let some n ← Nat.fromExpr? m | return .continue
+  let l := revRange n
+  --Convert the list to an `Expr`
```
b-mehta, line 235:
```suggestion
  -- Extract the natural number from the expression
  let_expr revRange m ← e | return .continue
  -- Recover the natural number as a term of type `Nat`
  let some n ← Nat.fromExpr? m | return .continue
  let l := revRange n
  -- Convert the list to an `Expr`
```

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+  return .visit <| Lean.toExpr l
+```
+
+**Pros**:
+* Requires writing a single simproc.
+* Assuming the type of the expression to be evaluated implements `ToExpr`, there is no need to reevaluate the expression in the meta world.
+
+**Cons**:
+* The expression to be evaluated is traversed twice: Once to create its evaluation, then once more in the typechecking of the proof by `rfl`.
```
b-mehta, line 244: The first one here is traversal by the vm, which shouldn't be that expensive at all; I'd guess negligible compared to the second.

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+A more general approach would be to manually construct the proof term we need to provide.
+In our case, we can do this in a recursive manner.
+```
+open Qq
+
+private theorem revRangeInduct {n : ℕ} {l : List ℕ}
+    (hl : revRange n = l) : revRange (n+1) = n :: l := by
+  induction n with
+  | zero => aesop
+  | succ n h => rw [←hl] ; rfl
```
b-mehta, line 259:
```suggestion
  | succ n h => rw [←hl]; rfl
```

posts/how-to-write-a-simproc.md
Paul-Lez: True! But only for the `n = 0` case - I figured never using definitional equalities would be a little cumbersome:)

b-mehta: It's cumbersome but given that definitional equalities are explicitly mentioned plenty of times here, and the Pro you give is explicitly that you're not relying on definitional equalities, I think it's pretty confusing to keep that! The fix would be to prove `revRange_zero`, and use that as the proof instead of `rfl`.

eric-wieser: Doing that probably result in a proof that type-checks ever-so-slightly faster!

posts/how-to-write-a-simproc.md
YaelDillies: Do you have any idea how to measure that?

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,370 @@
...
+  let l := revRange n
+  --Convert the list to an `Expr`
+  return .visit <| ToExpr.toExpr l
+```
+
+**Pros**:
+* Requires writing a single simproc.
+* Assuming the type of the expression to be evaluated implements `ToExpr`, there is no need to reevaluate the expression in the meta world.
+
+**Cons**:
```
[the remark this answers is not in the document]
eric-wieser: You can't literally evaluate the expression in the first place if it is not computable. Eg, if you want to compute the factors.

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,373 @@
...
+* Simplifying `revRange n` for a big input numeral `n` might involve a lot of simplification steps.
+  In this specific case, the number of simplification steps is linear in `n`.
+  Simplification steps matter because each of them increases the size of the proof term.
+* `revRange n` could find itself (partially) evaluated even if `n` isn't a numeral.
+  Eg `simp [revRange_zero, revRange_succ]` on `⊢ revRange (n + 3) = revRange (3 + n)` will result in `⊢ n + 2 :: n + 1 :: n :: revRange n = revRange (3 + n)`.
+  This is in general highly undesirable.
+
+## The definitional approach
+
+In cases where the evaluation is definitionally equal to to the original expression, one may write a dsimproc instead of a simproc.
```
b-mehta, line 226:
```suggestion
In cases where the evaluation is definitionally equal to the original expression, one may write a dsimproc instead of a simproc.
```

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,373 @@
...
+```
+
+**Pros**:
+* Requires writing a single simproc.
+* Assuming the type of the expression to be evaluated implements `ToExpr`, there is no need to reevaluate the expression in the meta world.
+
+**Cons**:
+* The expression to be evaluated is traversed twice: Once to create its evaluation, then once more in the typechecking of the proof by `rfl`.
+* The produced `rfl` proof could be heavy.
+* Only works when the evaluation and conversion back to an expression is definitionally equal to to the original expression.
```
b-mehta, line 248:
```suggestion
* Only works when the evaluation and conversion back to an expression is definitionally equal to the original expression.
```

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,373 @@
...
+In cases where the evaluation is definitionally equal to the original expression, one may write a dsimproc instead of a simproc.
+The syntax to declare a dsimproc is rather to simprocs, with a small difference: we now need to return a `Simp.DStep` instead of a `Simp.Step`; in practice this amounts to providing the expression our program has produced without providing the proof (indeed, this is just `rfl`!)
+
+To compute `revRange` using the dsimproc approach, we can do the following:
+```lean
+dsimproc_decl revRangeCompute (revRange _) := fun e => do
+  -- Extract the natural number from the expression
+  let_expr revRange m ← e | return .continue
+  -- Recover the natural number as a term of type `Nat`
+  let some n ← Nat.fromExpr? m | return .continue
```
b-mehta, line 235:
```suggestion
  let some n := m.nat? | return .continue
```
optional, but I think this is the preferred spelling of this now

posts/how-to-write-a-simproc.md
b-mehta: Actually it depends more precisely on what you mean by "create its evaluation" and "traversed". Does "expression to be evaluated" mean `revRange 10` or does it mean the created RHS?
But to answer your question, 
```
set_option trace.profiler true
set_option trace.profiler.threshold 0
```
gives you a bunch of profiling information.

b-mehta: (But also, questions of timing and complexity are a bit moot here since it breaks above n >= 250)

YaelDillies: I've just removed that ::

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,382 @@
...
+  -- Convert the list to an `Expr`
+  return .visit <| Lean.toExpr l
+```
+
+**Pros**:
+* Requires writing a single simproc.
+* Assuming the type of the expression to be evaluated implements `ToExpr`, there is no need to reevaluate the expression in the meta world.
+
+**Cons**:
+* The function needs to be computable to be evaluated in the meta world.
```
b-mehta, line 253: These two lines appear to contradict each other

posts/how-to-write-a-simproc.md
```diff
@@ -0,0 +1,382 @@
...
+    | 0 => ⟨q(([] : List ℕ)), q(revRange_zero)⟩
+    | n + 1 =>
+      let ⟨l, pf⟩ := go n
+      ⟨q($n :: $l), q(revRange_succ_eq_of_revRange_eq $pf)⟩
+  let ⟨l, pf⟩ := go n
+  return .visit { expr := l, proof? := pf }
+```
+
+**Pros**:
+* Works regardless of definitional equalities.
```
b-mehta, line 285: I'm not sure if you want to give more detail, but "manually construct the proof term" can be - in my eyes - quite a strong upside. For instance, the norm num extension for Nat.sqrt uses 
```
lemma nat_sqrt_helper {x y r : ℕ} (hr : y * y + r = x) (hle : Nat.ble r (2 * y)) :
    Nat.sqrt x = y := b
```
to construct the proof, the idea being that this is sufficient information to prove the square root computation, and both of these hypotheses are (relatively) easy to prove and type-check. (This is a norm num extension rather than a simproc, but I'm sure you can see that the same applies!) 

Now that I write this, I think the example you gave in the previous post of Nat.reduceDvd is an equally compelling example of this (we're not reducing `Dvd.dvd`, instead propositionally giving a sufficient condition for its truth/falsity), and doesn't require new explanation!

YaelDillies: Very good idea! I've done just that

posts/simp-made-simple.md
```diff
@@ -0,0 +1,153 @@
...
+  Continue transformation with the given expression (defaults to current expression).
+  For `pre`, this means visiting the children of the expression.
+
+  For `post`, this is equivalent to returning `done`. -/
+  | continue (e? : Option Expr := none)
+  deriving Inhabited, Repr
+```
+Note: The above snippet is a simplification and the constructors as shown actually belong to `Lean.TransformStep`, which `Lean.Meta.Simp.DStep` is an `abbrev` of.
+
+<span style="color:red">**(Yaël): Why is there a mismatch in docstrings between `Step.continue` and `DStep.continue`? [Zulip](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Simp.2EStep.2Econtinue.20vs.20Simp.2EDStep.2Econtinue/with/509056271)**</span>
```
Ruben-VandeVelde, line 132: Do you want to remove this?

posts/simp-made-simple.md
```diff
@@ -0,0 +1,151 @@
...
+A simproc takes in an expression and outputs a simplification step, possibly after modifying the current simp state (e.g. by adding new goals to be closed by the discharger).
+This behavior is formally encapsulated by the [`Simproc`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Meta.Simp.Simproc#doc) type:
+```lean
+abbrev Simproc  := Expr → SimpM  Step
+```
+
+`simp` does not consume bare elements of type `Simproc`.
+Instead, a simproc is an element of type `Simproc` annotated with tbe extra data mentioned in the overview subsection, like whether the simproc is `pre` or `post`.
+
+Similarly, there is a [`Simproc`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Meta.Simp.DSimproc#doc) type:
```
edegeltje, line 148:
```suggestion
Similarly, there is a [`DSimproc`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Meta.Simp.DSimproc#doc) type:
```

edegeltje (commented):
thx for writing this

posts/simp-made-simple.md
```diff
@@ -0,0 +1,151 @@
...
+All the `simp`-specific declarations introduced in this section are in the `Lean.Meta` or `Lean.Meta.Simp` namespace.
+
+## Overview
+
+When calling `simp` in a proof, we give it a *simp context*.
+This is made of a few different things, but for our purposes think of it as the set of lemmas/simprocs `simp` is allowed to rewrite with, namely the default simp lemmas/simprocs plus the lemmas/simprocs added explicitly minus the lemmas/simprocs removed explicitly.
+For example, `simp [foo, -bar]` means "Simplify using the standard simp lemmas/simprocs except `bar`, with `foo` added".
+
+A perhaps surprising fact is that every simp lemma is internally turned into a simproc for `simp`'s consumption.
+From then on, we will refer to simp lemmas/simprocs as *procedures*.
```
jcommelin, line 41:
```suggestion
From now on, we will refer to simp lemmas/simprocs as *procedures*.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,151 @@
...
+## `Simproc`s
+
+A simproc takes in an expression and outputs a simplification step, possibly after modifying the current simp state (e.g. by adding new goals to be closed by the discharger).
+This behavior is formally encapsulated by the [`Simproc`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Meta.Simp.Simproc#doc) type:
+```lean
+abbrev Simproc  := Expr → SimpM  Step
+```
+
+`simp` does not consume bare elements of type `Simproc`.
+Instead, a simproc is an element of type `Simproc` annotated with tbe extra data mentioned in the overview subsection, like whether the simproc is `pre` or `post`.
```
jcommelin, line 146: The section on Step is great, and contains a lot of detail. These latter sections seem somewhat rushed and the ending of the blogpost feels abrupt to me.
For example, this sentence about annotations with extra data... could that be accompanied by a code block that shows exactly what is going on?

And would it make sense to give the exact definition of the `simpM` monad, and add a bit more detail on it?

Maybe you could even end with a little cliff hanger about what will happen in the final blogpost?

Paul-Lez: Sounds good! I'm planning on spending some extra time adding stuff about `SimpM` to the post this coming month.

kim-em: 🏓  @Paul-Lez and @YaelDillies, it would be great to get this out! :-)

Paul-Lez: @kim-em this is now ready to review if you fancy taking a look;) 
If you're very busy, I think the first part is probably more ready/polished than the second one!

posts/simp-made-simple.md
```diff
@@ -0,0 +1,330 @@
...
+    usedTheorems : UsedSimps
+    numSteps     : Nat
+    diag         : Diagnostics
+  ```
+  This is something we can achieve using the `StateRefT` monad transformer, which takes as input a state type (`Simp.State` in our case) and a monad, and creates a new monad that can read _and write_ this state. In other words, `StateRefT Simp.State MetaM` is a souped up version of `MetaM` that can now track extra information by storing (and updating) at term of type `Simp.State`.
+
+3) The second monad transformer application: `ReaderT Simp.Context $ StateRefT Simp.State MetaM`. 
+  The `SimpM` monad should also be able to access the "context" that `simp` is running in, e.g. which simp theorems it has access to and so on. This is captured by the type `Simp.Context`.
+  Here, the situation is not quite the same as when we were adding a `Simp.State` state to `MetaM`: while we will often want to change the state during the `simp` call, the context should always be the same (in programmer lingo: _immutable_)
+  Thus, we use a different monad transformer called `ReaderT`, which is almost identical to `StateT`, but outputs a new monad that can only read the type passed as parameter.
```
Vierkantor: Mildly pedantic, but `ReaderT` can change the value in the calls it makes (e.g. [`ReaderT.adapt`](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#ReaderT.adapt) as equivalent to `modify` for a Reader). In programming language nerd terms, `State(Ref)T` gives you a global variable, and `ReaderT` gives you dynamic scoping.

(Does simp in practice use that though?)

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+  /-- Returns the result without visiting any subexpressions. -/
+  | done (r : Result)
+-- Note: the docstrings here are simplified versions of the real docstrings (which can be found in the Lean source code)
+```
+> If a procedure fails to simplify an expression, it should return `continue none`.
+  Both `visit` and `done` signify success.
+
+Whenever a simproc is called on a given expression, it outputs a `Step`, which determines what will happen next during the `simp` call. 
+Since every simproc call is running a metaprogram to produce the output `Step`, the constructor that ends up used may vary according to the input. 
+For example, a given simproc may in some cases use `visit` and in others use `continue`.
```
YaelDillies: This sentence is quite strange to me: it bases itself off the assumption that people expect that a simproc always uses the same `Step` constructor, but four lines above we explain how to use the constructors according to whether the simproc has failed or not. Could you repackage the info in a way that acknowledges the previous sentence?

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+  so this simproc uses `visit` to output the result of this simplification.
+
+## The `SimpM` monad
+
+In this section, we take a look at another key component of the internals of simp, namely the `SimpM` monad. 
+
+### `SimpM`
+
+[`SimpM`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Meta.Simp.SimpM#doc) is the monad that tracks the current context `simp` is running in (what `simp` theorems are available, etc) and what has been done so far (e.g. number of steps so far, theorems used).
+In particular this also captures the `MetaM` context.
```
YaelDillies:
```suggestion
In particular it also captures the `MetaM` context.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+
+Let's go through this in more detail. The monad `SimpM` is defined using monad transformers as follows:
+```lean
+abbrev SimpM := ReaderT Simp.MethodsRef $ ReaderT Simp.Context $ StateRefT Simp.State MetaM
```
YaelDillies:
```suggestion
abbrev SimpM := ReaderT Simp.MethodsRef <| ReaderT Simp.Context <| StateRefT Simp.State MetaM
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+1) The monad `MetaM`. This is one of the fundamental monads for metaprogramming in Lean. 
+  The state of `MetaM` allows one to access things like:
+    - Information about the file we're running in (e.g. name, imports, etc)
+    - Information about what definitions/theorems we're allowed to use
+    - What local variables/declarations we have access to
+
+2) The first monad transformer application: `StateRefT Simp.State MetaM`. 
+  The idea here is the following: since the goal of the `SimpM` monad is to track the state of a `simp` call
+  (i.e. what's happening, as the program runs), we need to capture more information than what `MetaM` gives us. 
+  Specifially, we want a monad that can track the state of what's happening via the following structure:
```
YaelDillies:
```suggestion
  Specifically, we want a monad that can track what's happening via the following structure: 
```
"the state" and "what's happening" seem to refer to the piece of information

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+    ```lean
+    structure Simp.State where
+      cache        : Cache
+      congrCache   : CongrCache
+      dsimpCache   : ExprStructMap Expr
+      usedTheorems : UsedSimps
+      numSteps     : Nat
+      diag         : Diagnostics
+    ```
+  This is something we can achieve using the `StateRefT` monad transformer, which takes as input a state type (`Simp.State` in our case) and a monad, and creates a new monad that can read _and write_ this state. In other words, `StateRefT Simp.State MetaM` is a souped up version of `MetaM` that can now track extra information by storing (and updating) at term of type `Simp.State`.
```
YaelDillies:
```suggestion
  This is something we can achieve using the `StateRefT` monad transformer, which takes as input a state type (`Simp.State` in our case) and a monad, and creates a new monad that can read _and write_ this state.
  In other words, `StateRefT Simp.State MetaM` is a souped up version of `MetaM` that can now track extra information by storing (and updating) a term of type `Simp.State`.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+
+3) The second monad transformer application: `ReaderT Simp.Context $ StateRefT Simp.State MetaM`. 
+  The `SimpM` monad should also be able to access the "context" that `simp` is running in, e.g. which simp theorems it has access to and so on. This is captured by the type `Simp.Context`.
```
YaelDillies:
```suggestion
  The `SimpM` monad should also be able to access the "context" that `simp` is running in, e.g. which simp theorems it has access to and so on.
  This is captured by the type `Simp.Context`.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+  Here, the situation is not quite the same as when we were adding a `Simp.State` state to `MetaM`: while we will often want to change the state during the `simp` call, we don't expect the context to need to change. In programmer lingo, the context should be _immutable_.
```
YaelDillies:
```suggestion
  Here, the situation is not quite the same as when we were adding a `Simp.State` state to `MetaM`:
  while we will often want to change the state during the `simp` call, we will never need to change the context.
  In programmer lingo, the context should be _immutable_.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+  Thus, we use a different monad transformer called `ReaderT`, which is almost identical to `StateT`, but outputs a new monad where one can only read the type passed as parameter. 
+    > For completness: when working with `ReaderT`, one can still locally overide the
```
YaelDillies:
```suggestion
    > For completness: when working with `ReaderT`, one can still locally override the
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+    > value of the variable that the monad keeps track of by using `withReader`. Intuitively, 
+    > the difference between `State(Ref)T` and `ReaderT` is the following: 
+    > - In `State(Ref)T`, one has access to a global variable that can be modified at 
+    >   will,
```
YaelDillies:
```suggestion
    > - In `State(Ref)T`, one has access to a global variable that can be modified at will,
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+  The `SimpM` monad should also be able to access the "context" that `simp` is running in, e.g. which simp theorems it has access to and so on. This is captured by the type `Simp.Context`.
+  Here, the situation is not quite the same as when we were adding a `Simp.State` state to `MetaM`: while we will often want to change the state during the `simp` call, we don't expect the context to need to change. In programmer lingo, the context should be _immutable_.
+  Thus, we use a different monad transformer called `ReaderT`, which is almost identical to `StateT`, but outputs a new monad where one can only read the type passed as parameter. 
+    > For completness: when working with `ReaderT`, one can still locally overide the
+    > value of the variable that the monad keeps track of by using `withReader`. Intuitively, 
+    > the difference between `State(Ref)T` and `ReaderT` is the following: 
+    > - In `State(Ref)T`, one has access to a global variable that can be modified at 
+    >   will, 
+    > - In `ReaderT`, given a program `x : ReaderT a m`, one can only chose to *execute* `x` with
```
YaelDillies:
```suggestion
    > - In `ReaderT`, given a program `x : ReaderT a m`, one can only choose to *execute* `x` with
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+    >   a different context. In particular, the context before and after the execution of `x` stays the same.
+
+4) The final monad transformer application: `ReaderT Simp.MethodsRef $ ReaderT Simp.Context $ StateRefT Simp.State MetaM`. 
+  This outputs a monad that has access to `Simp.Method` (passed via a ref). 
+  This capture the "pre" and "post" procedures that `simp` can use, as well as the discharger that `simp` can use, etc.
```
YaelDillies: This is ambiguous: do you mean `simp` the tactic, or `simp` a tactic call?
```suggestion
  This capture the "pre" and "post" procedures that a given `simp` call can use, as well as the discharger that a given `simp` call can use, etc.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+
+## `Simproc`s
+
+We can finally define what a `simproc` is formally.
```
YaelDillies:
```suggestion
Let's now formally define what a `simproc` is.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+    >   a different context. In particular, the context before and after the execution of `x` stays the same.
+
+4) The final monad transformer application: `ReaderT Simp.MethodsRef $ ReaderT Simp.Context $ StateRefT Simp.State MetaM`. 
+  This outputs a monad that has access to `Simp.Method` (passed via a ref). 
+  This capture the "pre" and "post" procedures that `simp` can use, as well as the discharger that `simp` can use, etc.
+
+## `Simproc`s
+
+We can finally define what a `simproc` is formally. 
+Recall that intuitively, a simproc takes in an expression and outputs a simplification step, possibly after modifying the current `SimpM` state (e.g. by adding new goals to be closed by the discharger).
```
YaelDillies:
```suggestion
Recall that, intuitively, a simproc takes in an expression and outputs a simplification step, possibly after modifying the current `SimpM` state (e.g. by adding new goals to be closed by the discharger).
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+This behavior is formally encapsulated by the [`Simproc`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Meta.Simp.Simproc#doc) type:
+```lean
+abbrev Simproc  := Expr → SimpM  Step
```
YaelDillies:
```suggestion
abbrev Simproc := Expr → SimpM Step
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+def mySimproc (e : Expr) : SimpM Step := do
+  -- Various manipulations involving the expression `e`
+  ...
+  let step : Step := ...
+  ...
+  return step
+```
+
+The picture above is however a slight oversimplification, as `simp` does not consume bare elements of type `Simproc`. 
+Instead, a simproc is an element of type `Simproc` annotated with the extra data mentioned in the overview subsection, like whether the simproc is `pre` or `post`, and what kind of expression it matches on.
```
YaelDillies: I must say it's confusing that a simproc isn't just an element of type `Simproc`. I expect people will read the whole paragraph before realising. Could we have a warning much earlier, or something like that?

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+
+```lean
+def printExpressions (e : Expr) : SimpM Step := do
+  Lean.logInfo m!"{e}"
+  return .continue
+
+-- declare the simproc
+simproc_decl printExpr (_) := printExpressions
+```
+The last line is needed to "declare" the simproc officially - this is where we specify information like whether this is a pre/post procedure, and what expression we're matching on (here, we match on the pattern `_`, i.e. on everything!).
```
YaelDillies:
```suggestion
The last line is needed to "declare" the simproc officially - this is where we can specify priority, whether this is a pre/post procedure and what expression we're matching on (here, we match on the pattern `_`, i.e. on everything!). 
```
We didn't mention the pattern on L.46-47. Should we?

YaelDillies (commented):
The first 140 lines flow really well! Let's make sure this is also the case for the other 200 lines after that.

posts/simp-made-simple.md
```diff
@@ -0,0 +1,343 @@
...
+  | continue (e? : Option Result := none)
+  /-- Returns the result without visiting any subexpressions. -/
+  | done (r : Result)
+-- Note: the docstrings here are simplified versions of the real docstrings (which can be found in the Lean source code)
+```
+> If a procedure fails to simplify an expression, it should return `continue none`.
+  Both `visit` and `done` signify success.
+
+Whenever a simproc is called on a given expression, it outputs a `Step`, which determines what will happen next during the `simp` call. 
+Since every simproc call is running a metaprogram to produce the output `Step`, the constructor that ends up used may vary according to the input.
```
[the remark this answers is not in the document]
Vierkantor:
```suggestion
Since every simproc call is running a metaprogram to produce the output `Step`, the constructor that ends up being used may vary according to the input.
```

posts/simp-made-simple.md
```diff
@@ -0,0 +1,330 @@
...
+  considered is simplifiable (e.g. the condition in the `if` statement). 
+  When this is not the case, the `continue` constructor signals to `simp` that it should 
+  *not* attempt to simplify the expression again using the same simproc to prevent 
+  the simplification procedure from looping around.
+  More generally, `continue` is used in most simprocs as the "default" output
+  produced when the simproc was not able to make any simplification.
+
+In the case where the two expressions `e` and `e'` are definitionally equal,
+one can actually describe a simplification step using a simple structure,
+namely `DStep` (where the "d" stands for "definitional").
```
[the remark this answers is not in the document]
Paul-Lez: Indeed - moved this elsewhere

posts/simp-made-simple.md
```diff
@@ -0,0 +1,337 @@
...
+  proof? : Option Expr := none
+  -- Note, `Result` currently has an extra `cache` field that is both deprecated
+  -- and irrelevant to our current discussion.
+```
+> The proof is optional:
+> If `proof?` is set to its default `none` value, the equality is assumed to be definitional.
+
+> The proof is allowed to be an arbitrary `Expr`,
+> i.e. nothing ensures that it actually is a proof that `e = e'`.
+> It is up to the author of the procedure to make sure that the generated proof terms are valid.
```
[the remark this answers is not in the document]
YaelDillies: My idea was to use the quote as an aparte

YaelDillies (commented):
This looks good now!

Vierkantor:
Alright! Let's get this merged 🎉

PR #113, opened 2025-06-26
Update and pin python dependencies

Ruben-VandeVelde:
This seems to work and I think pinning everything should ensure that it keeps working, but note that `pylock.toml` is a very new feature, so I could also see an argument for holding out on adopting it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,16 +15,18 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: install Python
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba # v6.3.1
         with:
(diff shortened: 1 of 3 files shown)


bryangingechen:
Thanks! I've been hoping to upgrade python and other dependencies in this project for a while. I'm assuming the instructions in the README are still accurate? (Maybe we want to add a `.python-version` and while we're at it move fully to a `uv`-based setup, but that might be a discussion for another time.)

Ruben-VandeVelde:
Oh, I missed the readme. I guess the `pip` command there might give you different versions of the dependencies, but then I don't know if we want to recommend `uv` to casual users.

.github/workflows/build.yml
```diff
@@ -15,16 +15,18 @@ jobs:
...
+        uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba # v6.3.1
         with:
-          python-version: 3.9
+          version: "0.7.15"
+          python-version: "3.13.5"
+          activate-environment: true
 
       - name: install Python dependencies
-        run: python -m pip install -r requirements.txt
+        run: uv pip install --preview -r pylock.toml
```
Vierkantor: Is there anything that will ensure `requirements.txt` and `pylock.toml` remain synchronized?

PR #139, opened 2026-08-22
blog on kernel bugs and junk values

siddhartha-gadgil:
My blog reflecting on trust and verification in Lean in the context of kernel bugs and junk values. This was originially posted at https://siddhartha-gadgil.github.io/automating-mathematics/posts/lean-kernel-bugs-and-junk-values/.

### DIFF
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -33,7 +33,7 @@ Any formal system, for humans or computers, is based on a formal language and lo
 
 For proof systems based on FOL a trusted kernel is indeed small - a few hours of work may be all it takes to write one. Simple logical systems also allow for powerful automation. However, writing even moderately complex mathematics in such a system requires an enormous amount of code.
 
-More complex logical systems are more *expressive* - allowing us to write more complex mathematics (or programs) reasonably concisely. However, verification is no longer so easy (and automation is also harder). There is a range of possible logical systems, including FOL, Higher Order Logic (HOL), and Calculus of Inductive Constructions (CIC). 
+More complex logical systems are more *expressive* - allowing us to write more complex mathematics (or programs) reasonably concisely. However, verification is no longer so easy (and automation is also harder). There is a range of possible logical systems, including FOL, Higher Order Logic (HOL), and Calculus of Inductive Constructions (CIC).
 
 Lean uses the Calculus of Inductive Constructions, which is expressive but complex. To make programming and proving easier (i.e., to increase expressivity), Lean adds features to CIC such as *proof irrelevance*, *quotients*, and *nested inductive types*.
 
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -23,7 +23,7 @@ Thus, individuals will have to be more cautious than one would ideally like, at 
 
 ## Trusted Kernels
 
-If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **AutoMath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
+If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **Automath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
 
 The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn principle*.
 
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -17,7 +17,7 @@ Reflections on how much confidence we can have in Lean verification, and what ha
 
 As *vibe coding* and AI generated proofs have proliferated, the Lean Prover has become a common tool to verify correctness. While it is undoubtedly advanced technology, Lean is not magic and cannot perfectly guarantee correctness. What Lean can do is allow a high degree of confidence after a modest amount of verification.
 
-This post is my reflections on how much confidence we can have in something verified by Lean, and what has to be checked to get this level of confidence. Roughly speaking our confidence depends on *trusting the kernel* and *checking definitions and statements*. These reflections are motivated by recent events (in August 2026) and discussions at a conference in early August 2026, which suggest that both these aspects are more complicated than one would ideally like.
+This post is my reflections on how much confidence we can have in something verified by Lean, and what has to be checked to get this level of confidence. Roughly speaking our confidence depends on *trusting the kernel* and *checking definitions and statements*. These reflections are motivated by recent events (in August 2026) and discussions at a conference in early August 2026, which suggest that both these aspects are more complicated than one would ideally like. There are additional (and more complex) issues with executable code, such as the compiler and the runtime, but I won't address them here (as of the time of writing we should not trust these levels).
 
 Thus, individuals will have to be more cautious than one would ideally like, at least at the present juncture, when trusting Lean proofs. There is also work to be done to raise the level of trust. Fortunately, it appears that there is willingness to do the work.
 
@@ -25,7 +25,7 @@ Thus, individuals will have to be more cautious than one would ideally like, at 
 
 If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **Automath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
 
-The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn principle*.
+The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn criterion*.
 
 ### Expressiveness, Verification, Automation
 
@@ -35,7 +35,7 @@ For proof systems based on FOL a trusted kernel is indeed small &mdash; a few ho
 
 More complex logical systems are more *expressive* &mdash; allowing us to write more complex mathematics (or programs) reasonably concisely. However, verification is no longer so easy (and automation is also harder). There is a range of possible logical systems, including FOL, Higher Order Logic (HOL), and Calculus of Inductive Constructions (CIC).
 
-Lean uses the Calculus of Inductive Constructions, which is expressive but complex. To make programming and proving easier (i.e., to increase expressivity), Lean adds features to CIC such as *proof irrelevance*, *quotients*, and *nested inductive types*.
+Lean uses the Calculus of Inductive Constructions, which is expressive but complex. To make programming and proving easier (i.e., to increase expressivity), Lean adds features in the kernel to CIC such as *nested inductive types*, and also introduces some axioms for *quotients* and *proof irrelevance*.
 
 Fortunately, it has been proved that these differently foundational systems are essentially equivalent (essentially because consistency of Lean assumes some large-cardinal axioms). However, Lean's kernel is not that small and is fairly complex - amounting to a few thousand lines of code (to the best of my knowledge). This is of course much smaller than Lean as a whole, and so there is indeed an enormous gain in terms of what needs to be verified. But verification of Lean's kernel is far from trivial.
 
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -77,7 +77,7 @@ It turned out that the code passed Nanoda because of an entirely different bug b
 
 The bug was immediately fixed in Lean. Over the next few days some more bugs were found and fixed.
 
-Many of these bugs involve strange meta-programming hacks whose intention is to clearly find and exploit the bugs, and others are not those that one is likely to encounter in normal Lean usage. So as far as normal Lean code is concerned there are no know kernel bugs. To increase confidence in one code, one can check (for instance via an LLM) whether strange meta-programming has been done or there is other non-idiomatic Lean usage. That said, once a few bugs are found something more has to be done to restore confidence.
+Many of these bugs involve strange meta-programming hacks whose intention is to clearly find and exploit the bugs, and others are not those that one is likely to encounter in normal Lean usage. So as far as normal Lean code is concerned there are no known kernel bugs. To increase confidence in one code, one can check (for instance via an LLM) whether strange meta-programming has been done or there is other non-idiomatic Lean usage. That said, once a few bugs are found something more has to be done to restore confidence.
 
 ### Verified kernels and *lean4lean*
 
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -23,7 +23,7 @@ Thus, individuals will have to be more cautious than one would ideally like, at 
 
 ## Trusted Kernels
 
-If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **Automath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
+If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **Automath** in the 1960s. The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
 
 The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn criterion*.
 
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -130,6 +130,6 @@ For individuals, I would recommend (as I have done) updating what you should che
 * One always needed to check statements.
 * One needs to check any new definitions, ideally by proving extra theorems.
 * In the case of standard definitions used directly or indirectly in the *statement*, be a bit careful if they use junk values. The more widely a definition is used the less likely it is that the error leaks, but a junk value is an error.
-* Slightly downgrade confidence due to the possibility of kernel bugs. For human written code, or AI written with active human involvement and understanding, I would personally reduce by only a tiny amount. For pure vibe coding: reduce by a little more if there is no strange metaprogramming, but if there is metaprogramming whose purpose is not clear be vary if this is an exploit.
+* Slightly downgrade confidence due to the possibility of kernel bugs. For human written code, or AI written with active human involvement and understanding, I would personally reduce by only a tiny amount. For pure vibe coding: reduce by a little more if there is no strange metaprogramming, but if there is metaprogramming whose purpose is not clear be wary if this is an exploit.
 
 For the Lean community, I hope we can build the understanding and tools to restore trust to the ideal level - as high a level as one can reasonably expect in the absence of magic.
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -65,7 +65,7 @@ However, `x` and `y` may not be definitionally equal, so we have to stop our sea
 
 In practice, this means that some additional criterion has to be introduced for giving up, and the behaviour of the type-checker depends on this. In the case of Trepplein (as implemented by Gabriel Ebner) there were actually configurable timeout parameters. Perhaps changing these would have allowed me to proceed. But at that time I got overwhelmed by the complexity and did not understand things as clearly as I do now, and so abandoned my efforts.
 
-I should emphasise that the undecidability is not a soundness issue - we only need an algorithm for type checking so that if it is accepted that `x` has type `A`, then indeed it does. If `x` has type `A` but the checker things it does not, then we will fail to prove something, weakening the prover. We will, however, not prove a false statement.
+I should emphasise that the undecidability is not a soundness issue - we only need an algorithm for type checking so that if it is accepted that `x` has type `A`, then indeed it does. If `x` has type `A` but the checker thinks it does not, then we will fail to prove something, weakening the prover. We will, however, not prove a false statement.
 
 ### Kernel bugs
 
--- a/posts/kernel-bugs-and-junk-values.md
+++ b/posts/kernel-bugs-and-junk-values.md
@@ -1,0 +1,135 @@
+---
+author: 'Siddhartha Gadgil'
+category: 'Design patterns'
+date: 2026-08-20T08:47:27+05:30
+description: ''
+has_math: true
+link: ''
+slug: KernelBugsJunkValues
+tags: ''
+title: Kernel Bugs and Junk Values
+type: text
+---
+
+Reflections on how much confidence we can have in Lean verification, and what has to be checked to get this level of confidence. These are prompted by spurious proofs that passed comparator due to kernel bugs, and errors occurring in formalisation due to junk values in the course of some major projects.
+
+<!-- TEASER_END -->
+
+As *vibe coding* and AI generated proofs have proliferated, the Lean Prover has become a common tool to verify correctness. While it is undoubtedly advanced technology, Lean is not magic and cannot perfectly guarantee correctness. What Lean can do is allow a high degree of confidence after a modest amount of verification.
+
+This post is my reflections on how much confidence we can have in something verified by Lean, and what has to be checked to get this level of confidence. Roughly speaking our confidence depends on *trusting the kernel* and *checking definitions and statements*. These reflections are motivated by recent events (in August 2026) and discussions at a conference in early August 2026, which suggest that both these aspects are more complicated than one would ideally like.
+
+Thus, individuals will have to be more cautious than one would ideally like, at least at the present juncture, when trusting Lean proofs. There is also work to be done to raise the level of trust. Fortunately, it appears that there is willingness to do the work.
+
+## Trusted Kernels
+
+If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **AutoMath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
+
+The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn principle*.
+
+### Expressiveness, Verification, Automation
+
+Any formal system, for humans or computers, is based on a formal language and logical rules. In the case of the "official" foundations of mathematics, *First-Order Logic* (FOL) gives the language and rules for deduction, with the axioms of *Set Theory* as the starting point for mathematics.
+
+For proof systems based on FOL a trusted kernel is indeed small - a few hours of work may be all it takes to write one. Simple logical systems also allow for powerful automation. However, writing even moderately complex mathematics in such a system requires an enormous amount of code.
+
+More complex logical systems are more *expressive* - allowing us to write more complex mathematics (or programs) reasonably concisely. However, verification is no longer so easy (and automation is also harder). There is a range of possible logical systems, including FOL, Higher Order Logic (HOL), and Calculus of Inductive Constructions (CIC). 
+
+Lean uses the Calculus of Inductive Constructions, which is expressive but complex. To make programming and proving easier (i.e., to increase expressivity), Lean adds features to CIC such as *proof irrelevance*, *quotients*, and *nested inductive types*.
+
+Fortunately, it has been proved that these differently foundational systems are essentially equivalent (essentially because consistency of Lean assumes some large-cardinal axioms). However, Lean's kernel is not that small and is fairly complex - amounting to a few thousand lines of code (to the best of my knowledge). This is of course much smaller than Lean as a whole, and so there is indeed an enormous gain in terms of what needs to be verified. But verification of Lean's kernel is far from trivial.
+
+### Multiple kernels
+
+Being aware of the danger of kernel bugs, Lean's creator Leo de Moura has long advocated for a way to strengthen the *de Bruijn principle* by having multiple independent kernels. There were three independent kernels for Lean 3 - one extracted from Lean, one written in Haskell, and one (trepplein) written in scala. With the advent of Lean 4 (the present version, and likely to be the version for many years to come) these no longer worked. So an independent kernel in Rust, *Nanoda*, was (commisioned and) implemented, along with documentation to write more kernels. Writing other kernels was facilitated with the *Lean kernel arena*.
+
+In addition, an important effort was *lean4lean* an implementation of Lean in Lean due to Mario Carneiro. This included an independent kernel. An effort is ongoing to verify correctness of the kernel.
+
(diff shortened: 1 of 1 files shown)


posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn principle*.
+
+### Expressiveness, Verification, Automation
+
+Any formal system, for humans or computers, is based on a formal language and logical rules. In the case of the "official" foundations of mathematics, *First-Order Logic* (FOL) gives the language and rules for deduction, with the axioms of *Set Theory* as the starting point for mathematics.
+
+For proof systems based on FOL a trusted kernel is indeed small - a few hours of work may be all it takes to write one. Simple logical systems also allow for powerful automation. However, writing even moderately complex mathematics in such a system requires an enormous amount of code.
```
grunweg: An em-dash is appropriate here (even before LLMs existed). I'm not sure about the correct syntax: do you simply do the following?
```suggestion
For proof systems based on FOL a trusted kernel is indeed small --- a few hours of work may be all it takes to write one. Simple logical systems also allow for powerful automation. However, writing even moderately complex mathematics in such a system requires an enormous amount of code.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+More complex logical systems are more *expressive* - allowing us to write more complex mathematics (or programs) reasonably concisely. However, verification is no longer so easy (and automation is also harder). There is a range of possible logical systems, including FOL, Higher Order Logic (HOL), and Calculus of Inductive Constructions (CIC).
```
grunweg: Same here.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+Fortunately, it has been proved that these differently foundational systems are essentially equivalent (essentially because consistency of Lean assumes some large-cardinal axioms). However, Lean's kernel is not that small and is fairly complex - amounting to a few thousand lines of code (to the best of my knowledge). This is of course much smaller than Lean as a whole, and so there is indeed an enormous gain in terms of what needs to be verified. But verification of Lean's kernel is far from trivial.
+
+### Multiple kernels
+
+Being aware of the danger of kernel bugs, Lean's creator Leo de Moura has long advocated for a way to strengthen the *de Bruijn principle* by having multiple independent kernels. There were three independent kernels for Lean 3 - one extracted from Lean, one written in Haskell, and one (trepplein) written in scala. With the advent of Lean 4 (the present version, and likely to be the version for many years to come) these no longer worked. So an independent kernel in Rust, *Nanoda*, was (commisioned and) implemented, along with documentation to write more kernels. Writing other kernels was facilitated with the *Lean kernel arena*.
```
grunweg: nit: do you lower-case Scala? (I always thought the opposite.) (Also below, several times.)

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+In addition, an important effort was *lean4lean* an implementation of Lean in Lean due to Mario Carneiro. This included an independent kernel. An effort is ongoing to verify correctness of the kernel.
```
grunweg:
```suggestion
In addition, an important effort was *lean4lean*, an implementation of Lean in Lean due to Mario Carneiro. This included an independent kernel. An effort is ongoing to verify correctness of this kernel.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+The bug involved nested inductive types, the same complex definitions that had made me sweat when trying to port trepplein. Talia Ringer has emphasised that this is not merely an issue with implementation - the type theory of nested inductive types is also not well understood.
+
+It turned out that the code passed Nanoda because of an entirely different bug but at the same place. Indeed this had been detected and fixed by the time the bug was announced. This part of the code in lean4lean was copied from Lean, with the bug being copied.
+
+The bug was immediately fixed in Lean. Over the next few days some more bugs were found and fixed.
+
+All of these bugs involve strange meta-programming hacks whose intention is to clearly find and exploit the bugs. So as far as normal Lean code is concerned there are no know kernel bugs. Indeed it is easy to check if strange meta-programming has been done. That said, once a few bugs are found something more has to be done to restore confidence.
```
grunweg: Have you checked "All of these bugs involve strange meta-programming hacks" carefully? Because I believe this is not quite true either. (The recent soundness bug about integer overflow uses no metaprogramming, for example.) What is still true is that a normal user will (almost surely) not accidentally hit this bug. The situation with AI and its incentives is... much harder to predict.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
```
grunweg, line 81: > Indeed it is easy to check if strange meta-programming has been done.

I think this statement deserves more nuance. It is easy to check if *no* metaprogramming has been done. Once AI also writes e.g. tactics... the situation is more complicated. There is no general solution for checking "does this Lean code behave badly" (that would probably amount to solving the halting problem, and there's also the "VW problem" --- Lean code could try to detect if it's being monitored).

Or are you referring to "we can run this through comparator"? (That only works if you have a well-defined challenge file, which works well for some projects and is harder for others.)

grunweg (commented):
Thanks for submitting this! I gave this another read - the changes you're making are clearly going in the right direction. I have some minor nits, and see a few more statements that are not quite accurate yet.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+Thus, individuals will have to be more cautious than one would ideally like, at least at the present juncture, when trusting Lean proofs. There is also work to be done to raise the level of trust. Fortunately, it appears that there is willingness to do the work.
+
+## Trusted Kernels
+
+If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **AutoMath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
+
+The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn principle*.
```
chenson2018: I have usually seen this in the literature as "de Bruijn criterion", is "principle" common in some contexts?

posts/kernel-bugs-and-junk-values.md
chenson2018: Another example is the mwe in https://gist.github.com/blume0/2f083ef60f9ccb8c961e888207e47ff7. There's no metaprogramming here.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+However, `x` and `y` may not be definitionally equal, so we have to stop our search at some stage. To get an algorithm we need to know how long we need to go on before giving up, or have a different half-algorithm to show inequality, or have some other conceptual approach. As proved in Mario Carneiro's thesis, there is no such algorithm.
+
+In practice, this means that some additional criterion has to be introduced for giving up, and the behaviour of the type-checker depends on this. In the case of trepplein (as implemented by Gabriel Ebner) there were actually configurable timeout parameters. Perhaps changing these would have allowed me to proceed. But at that time I got overwhelmed by the complexity and did not understand things as clearly as I do now, and so abandoned my efforts.
+
+I should emphasise that the undecidability is not a soundness issue - we only need an algorithm for type checking so that if it is accepted that `x` has type `A`, then indeed it does. If `x` has type `A` but the checker things it does not, then we will fail to prove something, weakening the prover. We will, however, not prove a false statement.
```
chenson2018: If undecidability is going to be discussed here, I think it should also be reflected in the discussion of external kernels. There are type theorists of the opinion that the value of external checkers is severely diminished in the presence of undecidable type checking. See https://mathstodon.xyz/@jonmsterling/116246759341442475 for instance. I think that even if the post decides to disagree on this point, this position should be discussed.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+type: text
+---
+
+Reflections on how much confidence we can have in Lean verification, and what has to be checked to get this level of confidence. These are prompted by spurious proofs that passed comparator due to kernel bugs, and errors occurring in formalisation due to junk values in the course of some major projects.
+
+<!-- TEASER_END -->
+
+As *vibe coding* and AI generated proofs have proliferated, the Lean Prover has become a common tool to verify correctness. While it is undoubtedly advanced technology, Lean is not magic and cannot perfectly guarantee correctness. What Lean can do is allow a high degree of confidence after a modest amount of verification.
+
+This post is my reflections on how much confidence we can have in something verified by Lean, and what has to be checked to get this level of confidence. Roughly speaking our confidence depends on *trusting the kernel* and *checking definitions and statements*. These reflections are motivated by recent events (in August 2026) and discussions at a conference in early August 2026, which suggest that both these aspects are more complicated than one would ideally like.
```
chenson2018: And perhaps the runtime too? This is qualitatively a bit different kind of trust, but has been a recent point of discussion...

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+Thus, individuals will have to be more cautious than one would ideally like, at least at the present juncture, when trusting Lean proofs. There is also work to be done to raise the level of trust. Fortunately, it appears that there is willingness to do the work.
+
+## Trusted Kernels
+
+If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **AutoMath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
```
chenson2018: I believe it is used styled as

```suggestion
If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **Automath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+Any formal system, for humans or computers, is based on a formal language and logical rules. In the case of the "official" foundations of mathematics, *First-Order Logic* (FOL) gives the language and rules for deduction, with the axioms of *Set Theory* as the starting point for mathematics.
+
+For proof systems based on FOL a trusted kernel is indeed small - a few hours of work may be all it takes to write one. Simple logical systems also allow for powerful automation. However, writing even moderately complex mathematics in such a system requires an enormous amount of code.
+
+More complex logical systems are more *expressive* - allowing us to write more complex mathematics (or programs) reasonably concisely. However, verification is no longer so easy (and automation is also harder). There is a range of possible logical systems, including FOL, Higher Order Logic (HOL), and Calculus of Inductive Constructions (CIC).
+
+Lean uses the Calculus of Inductive Constructions, which is expressive but complex. To make programming and proving easier (i.e., to increase expressivity), Lean adds features to CIC such as *proof irrelevance*, *quotients*, and *nested inductive types*.
```
chenson2018: In the context of kernel soundness, is it confusing to discuss choice and nested inductives together? It is perhaps worth clarifying what is in the kernel and what is an axiom since this is a common point of confusion.

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: Will fix.

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: Here I have said explicitly that these are my reflections. I don't think some of the objections are in good faith, and since this is a blog and not a paper/documentation, I don't feel compelled to acknowledge these points.

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: The runtime is of course an important point, as is the compiler. I don't want to address these but I should make it very clear that my scope is only kernel level proofs. Thanks.

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: Good point. I did not mention choice (as it is an opt in) but only proof-irrelevance and quotients, which are core features. But I should say that nested inductive types were added to the kernel and the other stuff was added as standard axioms. Both these can contribute to incosistencies though.

posts/kernel-bugs-and-junk-values.md
chenson2018: You are certainly not compelled to do so in a personal blog post. I won't speculate on the good faith of others. I personally find the technical argument of the interaction between undecidability and external checkers compelling, and I mention it in a good faith as something that concerns me as a Lean user.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+However, `x` and `y` may not be definitionally equal, so we have to stop our search at some stage. To get an algorithm we need to know how long we need to go on before giving up, or have a different half-algorithm to show inequality, or have some other conceptual approach. As proved in Mario Carneiro's thesis, there is no such algorithm.
+
+In practice, this means that some additional criterion has to be introduced for giving up, and the behaviour of the type-checker depends on this. In the case of Trepplein (as implemented by Gabriel Ebner) there were actually configurable timeout parameters. Perhaps changing these would have allowed me to proceed. But at that time I got overwhelmed by the complexity and did not understand things as clearly as I do now, and so abandoned my efforts.
+
+I should emphasise that the undecidability is not a soundness issue - we only need an algorithm for type checking so that if it is accepted that `x` has type `A`, then indeed it does. If `x` has type `A` but the checker things it does not, then we will fail to prove something, weakening the prover. We will, however, not prove a false statement.
```
mcdoll:
```suggestion
I should emphasise that the undecidability is not a soundness issue - we only need an algorithm for type checking so that if it is accepted that `x` has type `A`, then indeed it does. If `x` has type `A` but the checker thinks it does not, then we will fail to prove something, weakening the prover. We will, however, not prove a false statement.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+The bug involved nested inductive types, the same complex definitions that had made me sweat when trying to port Trepplein. Talia Ringer has emphasised that this is not merely an issue with implementation - the type theory of nested inductive types is also not well understood.
+
+It turned out that the code passed Nanoda because of an entirely different bug but at the same place. Indeed this had been detected and fixed by the time the bug was announced. This part of the code in lean4lean was copied from Lean, with the bug being copied.
+
+The bug was immediately fixed in Lean. Over the next few days some more bugs were found and fixed.
+
+Many of these bugs involve strange meta-programming hacks whose intention is to clearly find and exploit the bugs, and others are not those that one is likely to encounter in normal Lean usage. So as far as normal Lean code is concerned there are no know kernel bugs. To increase confidence in one code, one can check (for instance via an LLM) whether strange meta-programming has been done or there is other non-idiomatic Lean usage. That said, once a few bugs are found something more has to be done to restore confidence.
```
mcdoll:
```suggestion
Many of these bugs involve strange meta-programming hacks whose intention is to clearly find and exploit the bugs, and others are not those that one is likely to encounter in normal Lean usage. So as far as normal Lean code is concerned there are no known kernel bugs. To increase confidence in one code, one can check (for instance via an LLM) whether strange meta-programming has been done or there is other non-idiomatic Lean usage. That said, once a few bugs are found something more has to be done to restore confidence.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+Somewhat alarmingly, in Lean it is easy to prove that $1/0=0/0=0$. This is because division of natural numbers is defined on all natural numbers, so we have to define division by $0$. The choice was made to assign the value $0$ to such quotients. There are many such cases where Lean or Mathlib choose to assign a junk value to undefined quantities.
+
+Fortunately we cannot deduce $1=0$ from $1/0=0/0=0$ as the theorem $a/b * b = a$ needs the hypothesis $b \neq 0$. Indeed there is no danger that we can prove a false statement, as junk values do not make Lean unsound. What they do, though, is introduce a *semantic mismatch* - that a Lean definition does not coincide with what we think it means. One would like to say that since Mathlib definitions have been used in a lot of theorems, they are correct. But this is manifestly not the case - with junk values they are sometimes wrong. One instead has to make the rather slippery statement that Mathlib definitions are correct except in ways that will not affect anything you do - much less believable, and indeed not true sometimes as mentioned above.
```
mcdoll, line 116:
```suggestion
We cannot deduce $1=0$ from $1/0=0/0=0$ as the theorem $a/b * b = a$ needs the hypothesis $b \neq 0$. Indeed there is no danger that we can prove a false statement, as junk values do not make Lean unsound. What they do, though, is introduce a *semantic mismatch* - that a Lean definition does not coincide with what we think it means. One would like to say that since Mathlib definitions have been used in a lot of theorems, they are correct. But this is manifestly not the case - with junk values they are sometimes wrong. One instead has to make the rather slippery statement that Mathlib definitions are correct except in ways that will not affect anything you do - much less believable, and indeed not true sometimes as mentioned above.
```
No fortune involved ;-) (but feel free to ignore as well)

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+For important mathematical open problems there are repositories where the comparator statement has been formalized. In this case one simply uses that statement.
+
+### Standard definitions and Junk values
+
+All of the above depends on definitions that are in Mathlib being correct. Such definitions have been used in many theorems, and so errors are likely to have been detected. Except for one choice made in the development of Lean and Mathlib - the use of *junk values*. While this choice was made for understandable pragmatic reasons, it opens the way for errors. Indeed, as I learnt at the conference I mentioned from a talk by Sidhath Hariharan, such an error was made at one stage of the formalization of Viazovska's sphere packing theorem.
```
mcdoll, line 112: I think it would be better to write that a definition matches the informal one and not "definitions being correct". Definitions have no truth value, so they can't be correct or incorrect - they can match or not match what want them to be (as you explain below).

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Conclusions
+
+These episodes have made me slightly more cautious when trusting Lean code.
+
+For individuals, I would recommend (as I have done) updating what you should check with a Lean proof, and how much faith you have at the end of it:
+
+* One always needed to check statements.
+* One needs to check any new definitions, ideally by proving extra theorems.
+* In the case of standard definitions used directly or indirectly in the *statement*, be a bit careful if they use junk values. The more widely a definition is used the less likely it is that the error leaks, but a junk value is an error.
```
mcdoll, line 132: I actually don't really understand what you want to say. Mathlib has hundreds of theorems with junk values (a lot of calculus, where equalities for derivatives hold because if the function is non-differentiable, then both sides are zero. The point is that you can write `deriv` for any function but it only has meaning at the points where the function is differentiable).
Another example to think about is the current version of the Riemann hypothesis: the faithful version would be to exclude the pole of the zeta function at 1, but mathlib "uses" the fact that by coincidence the junk value is non-zero to not exclude the point. So it is not immediately obvious that it is equivalent to RH.

mcdoll (commented):
Thanks. Just read it and found a few typos.

I have two nitpicks, feel free to ignore.

posts/kernel-bugs-and-junk-values.md
SnirBroshi: This reads better to me, but I'm not sure
```suggestion
I should emphasise that the undecidability is not a soundness issue - we only need an algorithm for type checking so that if it is accepted that `x` has type `A`, then indeed it does. If `x` has type `A` but the checker things it does not, then we will fail to prove something, weakening the prover. We will not, however, prove a false statement.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+* Slightly downgrade confidence due to the possibility of kernel bugs. For human written code, or AI written with active human involvement and understanding, I would personally reduce by only a tiny amount. For pure vibe coding: reduce by a little more if there is no strange metaprogramming, but if there is metaprogramming whose purpose is not clear be vary if this is an exploit.
```
SnirBroshi: I think this is a typo
```suggestion
* Slightly downgrade confidence due to the possibility of kernel bugs. For human written code, or AI written with active human involvement and understanding, I would personally reduce by only a tiny amount. For pure vibe coding: reduce by a little more if there is no strange metaprogramming, but if there is metaprogramming whose purpose is not clear be wary if this is an exploit.
```
but it still sounds strange, so maybe
```suggestion
* Slightly downgrade confidence due to the possibility of kernel bugs. For human written code, or AI written with active human involvement and understanding, I would personally reduce by only a tiny amount. For pure vibe coding: reduce by a little more if there is no strange metaprogramming, but if there is metaprogramming whose purpose is not clear be wary of soundness exploits.
```

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: True, by design not fortune. But I would still use "fortunate".

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: This has been discussed on Zulip too. In my view, from the point of view of a Lean non-expert trying to use Lean to check a proof part of being correct is to be semantically aligned with standard mathematical terminology. So semantic misaligned is a case of wrong.

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: Again, I agree a subjective view but semantic misaligned I view as an error. In practice there have been wrong proofs by experts due to these and concerns in other cases (like the recent discussion on Anthropic's result about zeroes of zeta functions).

As for mathlib "using" the junk value, I would consider that an incorrect design choice.

mcdoll (approved):
I just want to flag I am happy with this blog post - I disagree with some of the things, but that is a matter for a discussion on zulip.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+Thus, individuals will have to be more cautious than one would ideally like, at least at the present juncture, when trusting Lean proofs. There is also work to be done to raise the level of trust. Fortunately, it appears that there is willingness to do the work.
+
+## Trusted Kernels
+
+If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **Automath** in the 1960s.The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
```
kim-em:
```suggestion
If we trust a computer system to test a proof, how do we know the system itself is correct? This fundamental question was addressed by de Bruijn while building his pioneering system **Automath** in the 1960s. The idea that he came up with was to separate a **kernel**, which checks correctness of proofs, from the rest of the system: the parts that help find proofs, provide interfaces etc. An analogue of the kernel is the part of a chess engine that checks whether moves are legal and whether we have a checkmate - much less code than the part that actually makes smart moves.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+The kernel has to be checked manually. However, if the kernel is relatively small, well-documented, and (these days) open sourced, one can hope that it has been checked by many people. Having such a kernel is now called the *de Bruijn criterion*.
```
kim-em, line 28: Why does it have to be checked manually? Isn't the whole point of `lean4lean` to checked it automatically?

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Expressiveness, Verification, Automation
+
+Any formal system, for humans or computers, is based on a formal language and logical rules. In the case of the "official" foundations of mathematics, *First-Order Logic* (FOL) gives the language and rules for deduction, with the axioms of *Set Theory* as the starting point for mathematics.
+
+For proof systems based on FOL a trusted kernel is indeed small &mdash; a few hours of work may be all it takes to write one. Simple logical systems also allow for powerful automation. However, writing even moderately complex mathematics in such a system requires an enormous amount of code.
```
kim-em, line 34: I'm not sure that "a trusted kernel" even makes sense for theorem provers without proof objects. Isabelle is just fundamentally different from Lean/Rocq.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+Fortunately, it has been proved that these differently foundational systems are essentially equivalent (essentially because consistency of Lean assumes some large-cardinal axioms). However, Lean's kernel is not that small and is fairly complex - amounting to a few thousand lines of code (to the best of my knowledge). This is of course much smaller than Lean as a whole, and so there is indeed an enormous gain in terms of what needs to be verified. But verification of Lean's kernel is far from trivial.
+
+### Multiple kernels
+
+Being aware of the danger of kernel bugs, Lean's creator Leo de Moura has long advocated for a way to strengthen the *de Bruijn principle* by having multiple independent kernels. There were three independent kernels for Lean 3 - one extracted from Lean, one written in Haskell, and one (Trepplein) written in Scala. With the advent of Lean 4 (the present version, and likely to be the version for many years to come) these no longer worked. So an independent kernel in Rust, *Nanoda*, was (commisioned and) implemented, along with documentation to write more kernels. Writing other kernels was facilitated with the *Lean kernel arena*.
```
kim-em, line 44: Here, and elsewhere, everything should be linkified, e.g. Nanoda, Lean kernel arena, etc.

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: Good point. Either it has to be checked manually or we need a certified kernel. So far as I know to date all kernels have only been checked manually, but if there are certified ones I should mention it. In any case I should make a correction.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+In addition, an important effort was *lean4lean* an implementation of Lean in Lean due to Mario Carneiro. This included an independent kernel. An effort is ongoing to verify correctness of this kernel.
+
+Thus, Lean code can be checked against multiple kernels. The hope is that this will not pass all of them if incorrect, as that will mean independent implementations have bugs in exactly the same place.
+
+Unfortunately, the chance of coinciding bugs is not as small as one would like. Firstly, part of *lean4lean* is not really independent as it borrows code from Lean itself. There is no such danger with a Rust implementation. However, bugs are most likely to appear in the most subtle parts of the foundations, so the locations of bugs will still be correlated.
```
kim-em, line 50: I mean, it's intentionally not independent. Initially Mario was trying to make it a direct translation of the official kernel! "bug for bug".

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: I will do this.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Trepplein times
+
+Trepplein is an independent type-checker for Lean 3 written by Gabriel Ebner in Scala. Since I knew Scala well and was also reasonably familiar with Lean-like foundations, I volunteered to try to port this to Lean 4. Unfortunately I did not finish this, but I learnt some things along the way. I should clarify that all this was some years ago, before we had AI chatbots, leave alone coding agents.
```
kim-em:
```suggestion
Trepplein is an independent type-checker for Lean 3 written by Gabriel Ebner in Scala. Since I knew Scala well and was also reasonably familiar with Lean-like foundations, I volunteered to try to port this to Lean 4. Unfortunately I did not finish this, but I learnt some things along the way. I should clarify that all this was some years ago, before we had AI chatbots, let alone coding agents.
```

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+Lean has an export format which is easy to parse. The code in Lean is exported in this format and independent type-checkers parse this and check correctness.
```
kim-em, line 56: Link to `lean4export`?

posts/kernel-bugs-and-junk-values.md
siddhartha-gadgil: I should clarify that some systems follow a design that allows for trusted kernels.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+In practice, this means that some additional criterion has to be introduced for giving up, and the behaviour of the type-checker depends on this. In the case of Trepplein (as implemented by Gabriel Ebner) there were actually configurable timeout parameters. Perhaps changing these would have allowed me to proceed. But at that time I got overwhelmed by the complexity and did not understand things as clearly as I do now, and so abandoned my efforts.
+
+I should emphasise that the undecidability is not a soundness issue - we only need an algorithm for type checking so that if it is accepted that `x` has type `A`, then indeed it does. If `x` has type `A` but the checker thinks it does not, then we will fail to prove something, weakening the prover. We will not, however, prove a false statement.
+
+### Kernel bugs
+
+The world now has AI systems that are phenomenally good at finding bugs. Using these, in early August 2026, a spurious Lean "proof" of the *Goldbach Conjecture*, a famous problem in mathematics, was posted, finding and exploiting a bug in the Lean kernel. This Lean proof also passed both Nanoda and lean4lean.
```
kim-em, line 72: Is this true? I thought it was some Collatz like statement?

siddhartha-gadgil: Oops. Indeed it was Collatz.

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+The bug was immediately fixed in Lean. Over the next few days some more bugs were found and fixed.
+
+Many of these bugs involve strange meta-programming hacks whose intention is to clearly find and exploit the bugs, and others are not those that one is likely to encounter in normal Lean usage. So as far as normal Lean code is concerned there are no known kernel bugs. To increase confidence in one code, one can check (for instance via an LLM) whether strange meta-programming has been done or there is other non-idiomatic Lean usage. That said, once a few bugs are found something more has to be done to restore confidence.
+
+### Verified kernels and *lean4lean*
+
+The best fix for the danger of kernel bugs is to have a kernel that is itself formally verified. The obvious candidate for this is *lean4lean*, on whose verification Mario has been working for some years. With the discovery of the kernel bugs many others have started contributing to this.
```
kim-em, line 84: Maybe mention Joachim's https://github.com/nomeata/lean-inductive-models, which transforms the export format to translate all of Lean's inductive types into things build out of only
`Eq`, `Nat`, `PUnit`, `PSigma'`, and `Quot` (at the expense of some definitional equalities becoming propositional). This could then be fed into a simpler kernel that didn't know about other inductives (or in particular nested inductives at all).

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+An attractive feature of these methods is that they can be automated. If an AI system has generated a statement and proof, it can independently generate corollaries to the statement, and subsequently prove them. A failure to prove the corollary indicates that the theorem is suspect, and this can then be re-examined.
+
+### Comparator
+
+In the case of many important theorems, the statement of the theorem only involves standard definitions, specifically those that are already in Lean's mathematical library Mathlib. If this is the case, the *comparator* model allows one to avoid making a mistake due to an error in new definitions. Namely, we have a separate file that only imports Mathlib and defines the statement of the problem. We then have a file that imports this statement file, imports the rest of the code of the project, and proves the statement.
+
+For important mathematical open problems there are repositories where the comparator statement has been formalized. In this case one simply uses that statement.
```
kim-em, line 108: "there are repositories" ... seems a bit unhelpful to the reader!

posts/kernel-bugs-and-junk-values.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Standard definitions and Junk values
+
+All of the above depends on definitions that are in Mathlib being correct. Such definitions have been used in many theorems, and so errors are likely to have been detected. Except for one choice made in the development of Lean and Mathlib - the use of *junk values*. While this choice was made for understandable pragmatic reasons, it opens the way for errors. Indeed, as I learnt at the conference I mentioned from a talk by Sidhath Hariharan, such an error was made at one stage of the formalization of Viazovska's sphere packing theorem.
```
kim-em, line 112:
```suggestion
All of the above depends on definitions that are in Mathlib being correct. Such definitions have been used in many theorems, and so errors are likely to have been detected. Except for one choice made in the development of Lean and Mathlib - the use of *junk values*. While this choice was made for understandable pragmatic reasons, it opens the way for errors. Indeed, as I learnt at the conference I mentioned from a talk by Sidharth Hariharan, such an error was made at one stage of the formalization of Viazovska's sphere packing theorem.
```

posts/kernel-bugs-and-junk-values.md
kim-em: "But this is manifestly not the case - with junk values they are sometimes wrong. "

Huh? This just seems wrong to me. :-) They don't mean what you might naively think they mean, but they aren't wrong. Maybe everyone else is wrong. 🤷‍♀️

